PR #3, opened 2017-07-03
Initial DOS v0.1 proposal

briandoconnor:
This is a proposal for a v0.1 DOS release based on the schema work of Brian W. and others from OHSU.  It also is informed by existing object storage systems such as:

* GNOS: http://annaisystems.com/ (as used by PCAWG, see https://pcawg.icgc.org)
* ICGC Storage: as used to store data on S3, see https://github.com/icgc-dcc/dcc-storage and https://dcc.icgc.org/icgc-in-the-cloud/aws
* HCA Storage: see https://dss.staging.data.humancellatlas.org/ and https://github.com/HumanCellAtlas/data-store
* the GDC Storage: see https://gdc.cancer.gov
* Keep by Curoverse: see https://arvados.org/ and https://github.com/curoverse/arvados

See the [README](https://github.com/ga4gh/data-object-schemas/blob/feature/protobuf-bdo-2/README.md) for more information.  The goals of DOS are to create a generic API on top of these and other projects, so workflow systems can access data in the same way regardless of project.  One section of the API focuses on how to read and write data objects to cloud environments and how to join them together as data bundles.  Another focuses on the ability to find data objects across cloud environments and implementations of DOS.

Please leave your comments here so we can move towards a v0.1 release soon.


briandoconnor:
Topics to discuss on the GA4GH call 7/3/2017:

* can we move the schemas into this repo? Easier to release since won't be tied to GA4GH schemas release.
* do we want to use the [data bundles concept](https://docs.google.com/document/d/1d-9eu5X6ioOlqOJ9kkY8lHvXDF-KoynlmqJbuKVPMF0/edit#heading=h.b3jd47oqdd2e)? Often times we want to be able to model related files (like a workflow output) together in some way.  The data bundle concept supports this.
* do we want to support versioning (of files and data bundles)?  Implicit support right now in the sense that you can get an array of files or data bundles and use timestamp to understand their version.
* do we want to support provenance?  Brian W's schema has a provenance object.
* do we want to include bio-specific data in the API?  This is a bigger question of how generic do we want DOS to be.  I lean towards really generic and, therefore, all biospecimen metadata should just be in JSON files that are part of the data bundle.  The alternative is to include links to other GDC/GA4GH metadata structures as Brian W. has done here.  
    * related, do we want to offer search on the bio-specific data?
* do all the timestamps need to be generated server-side?  This might be key for supporting versioning.
* questions about Brian W's schema:
    * how does Brian W. link datasets and files?  I don't see a `dataset_id` in the `DataObject` message.
    * how does Brian W. search on keys?  What field is `has_keys` searching on for DataObjects?  How should we expose search for other entities in the DataObjects or DataBundleObjects?

tetron:
A few initial thoughts:

* Are data object ids supposed to be globally unique?  Could they be URIs instead of opaque UUIDs?
* Would be more portable to use ISO 8601 timestamps instead of "milliseconds from epoch"
* Discourage using MD5 in 2017.  Recommend SHA256 or support multiple hashing algorithms. 
* Recommend JSON-LD for metadata
* Interest in using content addressing as a means of referencing files, eg query on hash field?
* Data bundles should be hierarchical
* "md5sum of file md5sums" -> not well specified
* id, dataset id, urls, alias -> what do these all mean?
* Does having several files in a data bundle imply they are also colocated, or can a data bundle refer to files hosted by different organizations?

briandoconnor:
Feedback for @tetron 's comments:

* Are data object ids supposed to be globally unique?  Could they be URIs instead of opaque UUIDs?
    * I think that's fine... I always assumed they would be globally unique IDs but the spec doesn't really say what they should be.  Do we need to define it?
* Would be more portable to use ISO 8601 timestamps instead of "milliseconds from epoch"
    * I think that's fine... HCA uses RFC3339, is that OK?
* Discourage using MD5 in 2017.  Recommend SHA256 or support multiple hashing algorithms. 
    * can we just say checksum?  And let implementors choose?
* Recommend JSON-LD for metadata
    * example of what that looks like?
* Interest in using content addressing as a means of referencing files, eg query on hash field?
    * can you make an endpoint for example?
* Data bundles should be hierarchical
    * interesting, I almost want to say it's up to the implementor?  What do you think?
* "md5sum of file md5sums" -> not well specified
    * OK, could make this more clear
* id, dataset id, urls, alias -> what do these all mean?
    * OK, good point, I'll try to clarify
* Does having several files in a data bundle imply they are also colocated, or can a data bundle refer to files hosted by different organizations?
    * I always assumed they are co-located

geoffjentry:
@briandoconnor @tetron I've heard people who have dealt with UUID based systems expressing a huge preference for URIs instead as the latter can provide some amount of semantic meaning. I do think its worth codifying either UUID or URI instead of "globally unique" as both have libraries available to handle.

re ISO8601 - IIRC RFC3339 isn't 100% ISO-8601 compliant but for our purposes effectively is. I find "ISO 8601" to be ambiguous anyways as while most people mean a specific format there are mutliple formats in there. IOW I think RFC3339 is AOK

bwalsh:
@briandoconnor I created some [slides](https://docs.google.com/presentation/d/1CCzsff4PZcNwysKPBzHSbTiP611HJO8jDVF1fjEY1nY/edit?usp=sharing) comparing this with UChicago's indexd service for monday's call

briandoconnor:
Thanks @bwalsh the slides look super helpful.  I think we should start the discussion with this and, depending on time, look at specific questions from this pull request.  It would be extremely helpful to translate your comparison and the feedback from the group today to changes to this branch before merging with develop.

briandoconnor:
So at this point, I think we need to resolve the following based on the comments above before a merge into develop.  We'll try to get through these on the 20170925, but, if not, I'll rely on comments here to close these out:

* take a look at Brian W's comparison to indexd/Signpost, can we adapt DOS to 1) include useful concepts from this and 2) make it easier to write a shim layer on top of Signpost for compatibility?

* discuss @geoffjentry and @tetron's comment on UUID vs. URL.  Do we envision DOS working like DNS where you have a UUID and some sort of decentralized collection of DOS instances that provide lookup?  Or should we go with a URL that can be resolved.  I think they are not incompatible with each other in the sense that we can define a URL structure for DOS servers, people can pass the UUIDs around as URLs, and, when talking to a given DOS instance, you can make requests using the UUIDs.  Thoughts?

* Checksums, should we require SHA256 across all DOS implementations?  Or let the implementor decide and advertise their checksum format?

* @tetron recommends JSON-LD for metadata.  So this is the contents of the `links` array, correct?  Do we have much say in this since Brian W is using JSON schema from GA4GH metadata project.  My understanding is they would have to adopt JSON-LD as their convention or am I not understanding the idea from Peter?

* @tetron asked for content addressing as a means of referencing files, is this now covered by the files-by-checksum endpoint?

* @tetron says "Data bundles should be hierarchical".  Let's get group feedback... I think they can be but this can simply be implemented by "/" in the name, right?

* are we happy with these optional elements:
    * aliases: alternative labels
    * urls: just what they say
    * links: to GA4GH objects for metadata
    * info: key/values
    * what about provenance, shouldn't it include a URL that includes the UUID + version (if supported?)

* Does having several files in a data bundle imply they are also colocated, or can a data bundle refer to files hosted by different organizations?

* we've got to think through how file write actually works. Right now you can effectively register a file but it's not clear how you register the cloud path, how you would actually upload the files bytes to DOS (or a signed URL), and how that cloud output location is communicated back. Specifically, say I just register a new file, how in my POST do I specify the 1) native cloud URLs or 2) how do I get a URL (signed or native) that I should then upload to. This is not at all clear.

* should /clouds become /locations?

* looks like the native URL and signed URL endpoints need version info in their requests.

* going with RFC3339

briandoconnor:
We primarily went through Brian W's slides on 20170925.  It generated some really great conversation on DOS, it's API, and what it should and should not do.

See https://docs.google.com/document/d/1kP7FzDJ9B7j7GMGeEiGZJazPLF4Z1p8K1a3FJazPigc/edit#

It includes a link to his slide pack too.

Some new questions that we need to discuss here . Good to pin down a few more topics that generated a bunch of great conversation today!:

* Metadata or no metadata intrinsic in the DOS API?  Do we allow metadata to be included in the DataObject or are these just files stored in the system?  Does DOS know anything about metadata?
* Common query interface in DOS?  Or another system?  Does DOS then allow you to search metadata if it’s stored in the system?  Like a generic search endpoint that all DOS systems support but not tied to particular metadata schema?  Or do we delegate to other systems that index content from a given DOS instance (or instances) and provide search capabilities?
* Federated or no?  When bringing up a DOS instance will it understand/support a federated network?  Will users be able to query across DOS instances to find data (by hash, UUID, etc?).  Or do we defer to other systems/groups for this?
* Eventing system in DOS?  If we don’t allow for queries of the metadata (or files storing metadata) in DOS it seems like we’d want a way to provide events.  This would allow, for example, other systems to automate indexing from DOS instances.  Otherwise, they would have to periodically walk the contents which is inefficient.

denis-yuen:
Does a common query interface imply that a common query language needs to be chosen and adopted across multiple implementors of DOS? We have a potentially similar challenge for TRS.

philloooo:
couple thoughts about previous comments, will have moooore later...
- UUID vs URL, semantic meaning is useful for actual analysis, but we need a permanent handle on this object even when this data object's location is moved or this data object is located in multiple places, I think it's mainly the decision made by a particular system about when to resolve an UUID to an URL, and pass it downstream processors. For example, when you want to process a file in a system that has resources in both AWS and GCP, the central manager API finds that it has available compute resources on AWS right now, it resolves an UUID to a list of URIs and pick the one on AWS, and send the url to the worker on AWS to process it.
- for hashes, I'd recommend to support multiple algorithms and this also gives flexibilities for new algorithms in the future. GDC [indexd](https://github.com/LabAdvComp/indexd/blob/master/indexd/index/schema.py#L35) currently support md5, sha1 and sha256, it's required for users to give at least one, but users can also give a combination of multiple.
- metadata support, I think generally a sophisticated system will have a more feature rich metadata API that supports a powerful query language, and the job of DOS is more about simply locating data in different storage systems. It's handy to have a simple key value metadata support, but make it optional. Because a lot of use cases will use a separate metadata API and we won't want to duplicate the info in this service.
- pubsub/webhooks support definitely will make the service more usable.

geoffjentry:
I'm still an enormous :-1: on UUIDs. Note that URI does *not* necessarily mean "www.foo.com/blahblah" so there's not necessarily any tie to a specific site. The comment that @philloooo made doesn't hold - a URI could be something like ga4gh://some/semantic/meaning and then DOS knows how to translate, just like it does with a UUID. The difference is that the URI might possibly be meaningful to someone reading it as a human instead of a UUID. Readability is a huge deal.

geoffjentry:
I do however agree with the point @philloooo made about metadata. I'd like to keep DOS slim & trim and not add on ten tons of stuff onto it. There can always be some other API which can tie in here.

tetron:
Agree with @geoffjentry about preferring URIs to UUIDs.  A URI format with an explicit scheme and a domain or namespace is vastly preferable to bare UUIDs.  A URI might still need to be resolved to a concrete URL for download, but there is a lot of value in being able to determine what type of identifier something is, and what organization issued it.

philloooo:
The UUID definitely shouldn't be passed around as plain, it would be `https://{one_id_provider}/{someprefix}/{UUID}` but just not the actual location URL. Does that make sense? @geoffjentry @tetron

philloooo:
Sorry about not making that clear, UUID as it's own is meaningless because you don't know who can resolve the UUID. For real cases that I imagine, the only thing that's good be used without context is hashes. The DOS at minimal would support UUID and hashes lookup. The users/services can either hold `https://{one_id_provider}/{someprefix}/{UUID}` if they want to talk with one id&data provider, or hold a hash that can be queried against multiple id&data providers.

philloooo:
One thing we discussed on 10/10/2017 meeting is immutable data objects and versioning.
From my point of view, the DOS should guarantee the immutability of the data, but you can move the data around and change annotation. In terms of actual fields, the core immutable fields for a data object would be:
```
- identifier(UUID)
- file_size
- hashes
- created
```
immutable, optional fields:
```
- version
```
mutable but core fields:
```
- updated
- urls
```
mutable metadata that I think could all be optional because they are not important for DOS core functionalities:
```
- file_name
- description
- info
- links
- alias
- ...
```
This satisfies most of the use cases which care about reproducibility. For some cases though, users want to get the latest version of a data object if available, so it might be helpful to have an identifier for MutableDataObject that's referenced by all versions of data objects?

geoffjentry:
@philloooo I'm not seeing what value the UUID brings in that context, at least as a required bit. What if I can uniquely describe my object in the `someprefix_or_none` portion? 

Instead I'd turn it around that these should be URIs where someone could jam a UUID in there to make it unique if they'd like to do that.

philloooo:
whether using UUID or user provided identifier is another topic then.  The prefix I put there was system controlled subdirectory. eg: `/files`.  Using  UUID instead of random user provided identifier is just simpler and much more systematically uniform.

geoffjentry:
Ok, then I'm definitely back to vehement disagreement. I can't imagine any world where requiring this as a UUID is superior to having a URI with user supplied information. They're both unique, they're both URIs as you said yourself. However requiring a UUID means that it's impossible to embed any semantic information in the URI. Using a user supplied URI means that one *can* use a UUID if they're lazy and want something simpler.

philloooo:
It isn't just about uniqueness but also about persistence. Semantic information may change over time, but Identifiers must remain persistent and should not change. you can always have semantic information in optional metadata fields and allow them to be mutable

philloooo:
See "Lesson 4" in http://journals.plos.org/plosbiology/article?id=10.1371/journal.pbio.2001414

geoffjentry:
Any information being put into the identifier should be immutable anyways, as you say, it won't change. 

I'm also :-1: on anything mutable about the file objects as IMO changing it makes it inherently a different object.

geoffjentry:
I'm just going to reiterate the request I had what was likely a year ago now that what I think DOS should be is something which provides S3 style URIs and a system which can translate between them to real URLs. Anything else belongs elsewhere.

geoffjentry:
damn, can't edit right now. Was going to add - the reason I say this is that obviously a hell of a lot of infrastructure can be built on that scheme (e.g. S3 itself, GCS, Azure Blob, etc). We should be striving for the simplest possible thing as that would enable the most potential users. If some subsets of users want more complicated things on top, those can be built.

I'm seeing this same general scheme pop up in a handful of consortiums and people are pointing back to this DOS effort. But the reason I'm hearing about it are that people are complaining that it's not what they want, that they want S3 style URIs. Granted I agree so they likely know they have a friendly ear.

philloooo:
So if you make semantic information immutable, and they become wrong information later, what do you do? for example if you had a typo in your filename and made this as your identifier, now it's published to DOS and will be there forever... For me,the key commitments of DOS should make are: 
1. identifiers persist, you don't delete ids once it's minted. You can't guarantee that you don't move data around, but given one identifier you can always get it's current location.
2. the content of the data should be immutable.

philloooo:
Agree it should be simpliest possible thing, that's why I prefer uuid over random user identifier. User defined identifiers can always be added on top of that, that's what we did, we just create user alias and join the bottom layer data objects by hashes.

allisonheath:
As someone who is probably more responsible for causing 👎 on UUIDs than most (UUID4 is just so great for distributed systems!) -- I agree in practice UUIDs are best left for system usage versus human usage. However, it's critical to make identifiers *opaque* and not contain semantic meaning. One good reference why is the paper @philloooo mentions (although kind of ironic that it's DOI encodes semantic information).

A bit more directly practical recommendation for ID generation is from Datacite (which has a lot of nice resources around digital identifiers in general): https://blog.datacite.org/cool-dois/. They also point out a number of reasons why they've moved away from DOIs containing semantic information while still striving to make the URI scheme human readable.

While you could make DOS more like DOI and allow people to make their own identifiers as long as they're unique and still have the same functionality, I think it would make it more robust to understand the community lessons learned around this and have a form of best practice minting baked in as part of the API/service.

geoffjentry:
Don't get me wrong, I understand the reasons against doing what I'm asking for, for instance I've had to deal with too many bozos who put PHI directly in file names. My point is to not throw the baby out with the bath water here. There are legit reasons why embedding some bits of completely immutable metadata become far easier for sanity sake.

Again, there is no reason the URI approach can't use UUIDs but the UUID approach but the inverse is not true.

allisonheath:
Yeah - I think I just wanted to point out that at least as coming into this conversation fairly late, it seems like everyone's in agreement about a URI approach (and the UUIDs might have been a bit of a red herring).

Looking at the DOI example - I could see a reasonable strategy that allows semantics in the prefix for high level items (e.g. institution), but then the final suffix is opaque so the more detailed aspects that are more likely to change aren't embedded.

philloooo:
@allisonheath that makes sense to me. @geoffjentry can you give some example of useful metadata to be embedded?

geoffjentry:
I was thinking about this more and I'm thinking that this argument stems from there being (at least) two different views of what DOS *is* in the first place. 

I know that myself and a few others (Not sure if this is where @tetron 's disagreement stems from) have been viewing this as a generic replacement for S3/GCS/Blob/OCS/etc. I believe that @philloooo and others (likely @briandoconnor and @allisonheath ) are viewing this as purely a generic API for standing in front of large data repositories. Further, I believe this is where the fundamental disconnect on what is "the simplest" is coming from.

In the latter case I do agree that preventing semantic information makes a lot more sense but in the former case the answer to the question is @philloooo is that it doesn't need to be justified as there are all sorts of reasons I might want to name a file something.

allisonheath:
@geoffjentry Ahhh that helps give some context! I think I was just going off of the original set of examples from @briandoconnor as those are all pretty much large data repositories, and thus made the assumption we were talking generic API in front of them.

Conversations of what something *is* are typically a better done in real time - so perhaps next week's meeting will be a chance to get on the same page there to get this PR wrapped up.

geoffjentry:
@allisonheath We *do* have a whole day to chat about the cloud workstream coming up in Orlando but I'm not sure who is/isn't going to be there.

I think it's always been known that there were multiple not-quite-the-same goals from the start (IIRC there's a doc somewhere where people wrote down what they wanted as these sorts of arguments kept erupting on calls) - I originally pitched what became TES because a) We (Cromwell team) *really* liked Google Pipelines API, b) we wanted to support more clouds, and c) If clouds supported the same API for a PAPI-like thing it'd make our lives super easy. While `c` hasn't quite happened yet it's actually possible (I know there are multiple cloud vendors considering it).

Likewise another thing Cromwell does a lot of is file I/O to the execution backend, so again, we need to talk to GCS, S3, etc ....... so there's a similar desire of if we can just write it once that's a lot of saved effort and less code to maintain.

I'm sure that's completely different than how e.g. @philloooo is picturing using DOS.

philloooo:
oh yeah that's completely different than what I thought. I consider this is just data DNS, instead of a generic API that reverse proxies to S3/GCS/etc.
btw, both Allison and me will go there

bwalsh:
Good discussion.   A reverse proxy to S3, etc.  including file upload / download is fairly ambitious.  Is it reasonable to assume a likely outcome is that: a) clouds will align around a de-facto standard (s3) or b) a separate cross platform/cross cloud solution will emerge that will be industry wide (not just cancer/informatics/etc).  If these assumptions are incorrect and we do want a cross cloud reverse proxy, wouldn’t a “data DNS” be a first step?
I'll be in Orlando as well.  Looking forward to meeting everyone face to face

david4096:
I'm excited to join such a lively conversation! I would very much like to move this to a working interinstitutional prototype in the minimum amount of time and these are what I see as blockers.

Metadata, although crucial for bioinformatics analysis, should not be a key feature of the DOS. Files may have metadata about them, but it should not attempt to be both a Object store and a metadata index about the contents of that object store. Having a metadata index over an Object store is of utility, but as @philloooo mentions, is not a core function of DOS.

My understanding of the core functionality is being able to address, via content and aliases, files in multiple underlying stores.

For this reason I look to public cloud stores and the minimum features that made them successful. S3 let's me quickly and easily syndicate files off my local computer to amazon. Rediscovering the contents of those files is up to me and managing useful metadata indexes over these files.

DOS should allow one to easily create clients that will syndicate to other clouds, or reason about the file via content (hash), alias, or identifier.

In addition to this, we want to offer a simple, single level way of describing files together. Files can be indexed and reindexed in data bundles without incurring new storage costs. My hope is that by constraining the metadata discussion in this way we can simply specify the JSON-LD for a metadata JSON file that clearly states which ontologies it uses. If the metadata changes the bundle changes!

I realize that there has been a lot of discussion on this and I want to keep on the table the possibility of supporting "show me files for this sample" type queries! I think that leads to lots of interesting use cases, which we can approach at a later date.

However, I consider that the concern of a different sort of index. DOS, to first order, simply needs to provide a way to abstract the simplest cloud storage operations. Even for a store that maintains millions of samples, the whole DOS index should be in the megabytes. Constraining that now and making it clear that we will not enforce metadata schemas to first order will improve uptake.

### Identifiers

For files, we have the benefit of content addressing, which, if used properly will be a huge boon for improving deduplication and data reuse.

However, the problem of ascribing useful identifiers and guaranteeing that someone is getting the file they expect, isn't easy. IDs, in the end, will be server local even if you ask people to generate intermediate UUIDs to identify their data. Providing the feature of "find by checksum" provides a nice balance to allow clients to reason about data in multiple DOS stores even if it hasn't been labeled similarly.

So, my suggestion is to try to allow people to describe their data in the ways they do using a list of identifiers or URLs as suits them. As long as we can construe a client to access the file based on what they provide, we should be OK. I would discourage folks making their own `protocol://` as s3 did unless they have a specific client. We want to play up the, easy, over-the-top nature of DOS and I think that http does just fine. 

+1 to data DNS. We can easily write data crawlers that can syndicate to multiple stores and pipelines that can grab data.

I hope we can move discussion to individual issues as quickly as possible! I took the work from this PR and removed some of the standing discussion items so we can start with an implementable schema. https://github.com/ga4gh/data-object-schemas/pull/4

geoffjentry:
@philloooo in particular - if it wasn't obvious, the lack of sound of me in the last few days was the sound of me giving up :P For the purposes of a "data DNS" (to use the awesome term @philloooo  made) and that only, my stances change radically.

philloooo:
@geoffjentry great that we are one step closer to consensus! "data DNS" is how people think about our ID service in CDIS :p

PR #4, opened 2017-10-13
DOS minimal schemas 

david4096:
Hi there! In hopes of stimulating the discussion towards a quick first useful iteration I offer this interpretation of the standing work. I removed optional fields that supported metadata features that seemed extraneous to the core functionality of abstracting cloud storage providers.

I realize I'm joining this conversation after it has started but I hope to be able to help everyone rally around a minimum set of features we would all find useful!

I'm hoping that with the additions of the Travis test and better requirements we might merge this PR and then create new PRs to discuss the given TODO items.

* DONE _TODO_ remove dataset_id, remove submitter_id


proto/data_objects.proto
```diff
@@ -0,0 +1,268 @@
...
+  repeated string urls = 8;
+
+  //OPTIONAL
+  string description = 9;
+
+  //OPTIONAL
+  repeated string aliases = 10;
+
+  //OPTIONAL
+  google.protobuf.Struct info = 12;
```
bwalsh: We should deprecate this.  Legacy from ga4gh schemas.

proto/data_objects.proto
```diff
@@ -0,0 +1,268 @@
...
+
+  // OPTIONAL the id of the file as submitted
+  string submitted_id = 14;
```
bwalsh: We should deprecate this. alternate ids should be in alias

proto/data_objects.proto
```diff
@@ -0,0 +1,268 @@
...
+  string dataset_id = 7;
+
+  //OPTIONAL
+  repeated string urls = 8;
+
+  //OPTIONAL
+  string description = 9;
+
+  //OPTIONAL
+  repeated string aliases = 10;
```
bwalsh: Do we need an alias of the bundle? Wouldn't one just create a second bundle?

proto/data_objects.proto
```diff
@@ -0,0 +1,268 @@
...
+  // REQUIRED
+  // Timestamp of update in RFC3339, identical to create timestamp in systems
+  // that do not support updates.
+  google.protobuf.Timestamp updated = 4;
+
+  //R EQUIRED
+  // A string representing a version, some systems may use checksum, a RFC3339
+  // timestamp, or incrementing version number. For systems that do not support
+  // versioning please use your update timestamp as your version.
+  string version = 5;
```
bwalsh: How exactly do we have a version? The what was the timestamp of the previous version? What was its hash value? Should we probably have a separate `DataObject repeated versions`  to hold different versions?

proto/data_objects.proto
```diff
@@ -0,0 +1,268 @@
...
+
+  //REQUIRED, the bundle SHA256 checksum which is a SHA256 checksum of all the
+  // SHA256 checksums of the files that make up this bundle
+  string checksum =  6;
+
+  //OPTIONAL
+  string dataset_id = 7;
+
+  //OPTIONAL
+  repeated string urls = 8;
```
bwalsh: Is this the set of all urls for the bundle?

proto/data_objects.proto
```diff
@@ -0,0 +1,268 @@
...
+  // REQUIRED
+  // Timestamp of update in RFC3339, identical to create timestamp in systems
+  // that do not support updates.
+  google.protobuf.Timestamp updated = 5;
+
+  // REQUIRED
+  // A string representing a version, some systems may use checksum, a RFC3339
+  // timestamp, or incrementing version number. For systems that do not support
+  // versioning please use your update timestamp as your version.
+  string version = 6;
```
bwalsh: see comment above re. version

proto/data_objects.proto
```diff
@@ -0,0 +1,268 @@
...
+  string checksum =  8;
+
+  // OPTIONAL
+  repeated string urls = 9;
+
+  // OPTIONAL
+  string description = 10;
+
+  // OPTIONAL
+  repeated string aliases = 11;
```
bwalsh: I think we need to add a separate Alias struct

Alias

metadata<string,string>
// we also need to account for https://github.com/ohsu-comp-bio/indexd/blob/master/indexd/alias/schema.py#L12-L41

proto/data_objects.proto
```diff
@@ -0,0 +1,268 @@
...
+  // A string representing a version, some systems may use checksum, a RFC3339
+  // timestamp, or incrementing version number. For systems that do not support
+  // versioning please use your update timestamp as your version.
+  string version = 6;
+
+  // OPTIONAL
+  string mime_type = 7;
+
+  // REQUIRED, the file SHA256 checksum
+  string checksum =  8;
```
bwalsh: indexd supports multiple checksum types. need to resolve
https://github.com/ohsu-comp-bio/indexd/blob/master/indexd/index/schema.py#L35-L76

proto/data_objects.proto
```diff
@@ -0,0 +1,268 @@
...
+  repeated string urls = 9;
+
+  // OPTIONAL
+  string description = 10;
+
+  // OPTIONAL
+  repeated string aliases = 11;
+
+  //OPTIONAL
+  google.protobuf.Struct info = 13;
```
bwalsh: deprecate ?

proto/data_objects.proto
```diff
@@ -0,0 +1,268 @@
...
+
+  // OPTIONAL the id of the file as submitted
+  string submitted_id = 15;
```
bwalsh: deprecate?

proto/data_objects.proto
```diff
@@ -0,0 +1,268 @@
...
+  // timestamp, or incrementing version number. For systems that do not support
+  // versioning please use your update timestamp as your version.
+  string version = 5;
+
+  //REQUIRED, the bundle SHA256 checksum which is a SHA256 checksum of all the
+  // SHA256 checksums of the files that make up this bundle
+  string checksum =  6;
+
+  //OPTIONAL
+  string dataset_id = 7;
```
bwalsh: Can we document this?  Unclear why/when I would use DataBundleObject.dataset_id  vs DataBundleObject.id

proto/data_objects.proto
david4096: Yeah, I think we need to use a data structure that combines an enum (checksum used) with the actual checksum.

proto/data_objects.proto
david4096: I think we should remove mentions of dataset id, it confuses the issue and adds a new concept which is out of scope for this API. This would also imply removing the search by `dataset_id` from the ListDataBundleObjects request.

proto/data_objects.proto
bwalsh: map<string, string> system_metadata = N;
map<string, string> user_metadata = N;

proto/data_objects.proto
david4096: This can be removed, The `DataObjects` all resolve to URLs, which the client will use.

proto/data_objects.proto
david4096: Added.

proto/data_objects.proto
bwalsh: The Alias should also have it's own hashes.

```
Hi Phillis.  Quick question:  In indexd, why does the alias have a its own `hashes`.  Is this a duplicate of the hashes in `index`? Or, does it allow the user to maintain their own, separately computed hash?

phillis [11:16 AM] 
It's like the foreign key to index document

[11:17] 
when you say I have this identifier for this data, you post to /alias and specify the hashes, then the id service should join the actual index document to find where the object is given the checksum

[11:18] 
basically we always trust the checksum to be cross referencable
walsbr [11:18 AM] 
Ok.   thanks. So, when the user posts the `alias` the if index not found, it is an exception.
```

proto/data_objects.proto
david4096: I'm imagining this would support having a single index entry for a bundle that says the GDC calls it "bundleX" and UCSC calls it "bundleY". Although two bundle messages supports the feature, for the purposes of resolving a bundle to kick off a workflow having a single message seems more clear.

bwalsh: ok.  thanks for clarifying.  lets add that as a comment

david4096:
Created some issues that will likely take some documentation and discussion to close (out of scope of this PR).

https://github.com/ga4gh/data-object-schemas/issues/12
https://github.com/ga4gh/data-object-schemas/issues/11
https://github.com/ga4gh/data-object-schemas/issues/10
https://github.com/ga4gh/data-object-schemas/issues/9

proto/data_objects.proto
david4096: https://github.com/ga4gh/data-object-schemas/issues/12

proto/data_objects.proto
```diff
@@ -0,0 +1,268 @@
...
+
+///-------------------------------------------------------------------
+// Resource: a file, API or other resource
+///-------------------------------------------------------------------
+message DataObject {
+  //REQUIRED
+  string id = 1;
+
+  //REQUIRED
+  string file_name = 2;
```
philloooo: can this be optional? data object doesn't necessarily need a name, or doesn't need to store the name in the dataobject

david4096: I advise removing filename and using Alias for everything.

Downloading files by filename is risky if two people name different files similarly. They will overwrite files unintentionally.

philloooo: removing filename will be great

bwalsh: +1 remove file name

proto/data_objects.proto
philloooo: if you use timestamp/checksum for version this should be optional?

bwalsh:
+1 great conversation and effort. thanks  @philloooo  @david4096

david4096:
@philloooo If you have a moment to take a look and give your cautious +1 that this is a step forward that would be great!

We have a few issues to solve before I am really confident we should push ahead with an implementation. https://github.com/ga4gh/data-object-schemas/issues/13

philloooo:
[this](https://github.com/ga4gh/data-object-schemas/pull/4#discussion_r145196614) is my last comment :)

proto/data_objects.proto
david4096: Yes, I think so.

david4096:
Thanks @philloooo I modified that field to be `optional`.

We need to document the guidelines for how versioning will work with syndication given your suggestions. I can imagine removing this field from the schema with a little more discussion.

Let's merge? and please add your perspectives on the standing problems in the issues! Thanks!

PR #5, opened 2017-10-15
Change all references from "file" to "object"

delagoya:
Propose to remove the word "file" from the specification and replace it with "object".


bwalsh:
+1 sounds good to me

david4096:
This @delagoya I think I got them all in #4 .

PR #24, opened 2017-10-30
Minimal schemas, test client/server

### LINKED ISSUE #7: Remove mentions of "clouds"
Thanks @delagoya

### LINKED ISSUE #9: Versioning
We need to better constrain how versioning is expected to be managed in practice.

Via @bwalsh

> How exactly do we have a version? The what was the timestamp of the previous version? What was its hash value? Should we probably have a separate DataObject repeated versions to hold different versions?

https://cloud.google.com/storage/docs/key-terms#immutability
http://docs.aws.amazon.com/AmazonS3/latest/dev/ObjectVersioning.html

> Google in particular has a version for object and a metaversion for metadata changes.
https://cloud.google.com/storage/docs/object-versioning#details
> This differs from amazon:
> Each Amazon S3 object has data, a key, and metadata. Object key (or key name) uniquely identifies the object in a bucket. Object metadata is a set of name-value pairs. You can set object metadata at the time you upload it. After you upload the object, you cannot modify object metadata. The only way to modify object metadata is to make a copy of the object and set the metadata.

> S3 breaks meta data into two arrays, system and user
http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html#object-metadata

### LINKED ISSUE #12: Checksum types
@bwalsh
> indexd supports multiple checksum types. need to resolve
https://github.com/ohsu-comp-bio/indexd/blob/master/indexd/index/schema.py#L35-L76

@david4096
> Yeah, I think we need to use a repeated data structure that combines an enum (checksum used) with the actual checksum. For SHA256 I think we mean the hexlified message digest, not the actual bytes.

This also suggests adding a feature to the request for specifying the type of checksum used, however, given the uniqueness of checksums I think it's not necessary. Could these other typed checksums be simply called another metadata key?

### LINKED ISSUE #13: Remove extraneous methods
There are currently proposed a few messages and endpoints that can be solved in a more general manner (`/clouds`, `/status`).

To close this, we will remove the methods that can currently be solved using simpler means.

Related to number #7

### LINKED ISSUE #15: Map of array of strings
The added `system_metadata` and `user_metadata` fields allow for tagging of strings to key values, but a more 'robust' model may be to have a map to a repeated string. Synapse.org uses that model for their object properties. 

So that 
https://github.com/ga4gh/data-object-schemas/blob/7858f2005944d8bb73e2e5c7f2447c3c459e625d/proto/data_objects.proto#L88

would be something like
` map<string, google.protobuf.ListValue> system_metadata = 12; `

### LINKED ISSUE #16: DataObjectBundle:  `updated` is Required?
https://github.com/ga4gh/data-object-schemas/blob/develop/proto/data_objects.proto#L23

Should be optional?

david4096:
Hi everyone! This PR provides a second pass on these schemas!

It includes a simple server in `app.py` and a client in `client.py`. They use the connexion and bravado libraries respectively to map the swagger functions to methods and python objects. Note that you have to run the `tools/prepare_swagger.sh` to make some last minute modifications to the swagger file to make it all work!

Run `app.py` and see `demo.py` and `gdc_dos.py` for examples of how one might work with the service defined in the schemas! Mind you, the server is meant for experimentation/testing purposes only!

Main concepts worth noting: 

* The `url` attribute is now searchable, using the ListDataObjectsRequest, as we believe that to satisfy some edge cases this will be necessary. Namely, to handle delete messages sent from blob stores may not contain the checksum of the file. 
* The `url` is also now a complex object that can offer its own metadata. This supports the use case of separate blob stores housing differing metadata.
* Checksums are repeated messages to support when more than when checksum is provided.
* The default checksum provided by many blob stores is md5, so the default has been changed to reflect that.

The checksum message includes an open string field to specify which checksum method was used (as opposed to enum).

Note: the go-swagger pipeline is opinionated about representing int64 as string in JSON, which JSONSchemas disagrees with. This leads to validation and serialization problems, so a patched bravado library is being used (see `requirements.txt`).

I will continue to improve the documentation and infrastructure surrounding working on the schemas. Thanks everyone who has helped get ideas into this! This PR should close #13 close #15 close #12 close #9 close #7 close #16 .

Check out DEVELOP.md https://github.com/david4096/data-object-schemas/blob/d600291cfc43a6bd7feace7bd9734fee9cff9c1c/DEVELOP.md

You can run your own swagger-ui using `app.py` http://localhost:8000/ui!

https://david4096.github.io/data-object-schemas/


proto/data_objects.proto
```diff
@@ -1,65 +1,65 @@
...
 message DataObject {
   // REQUIRED
+  // An idenitifier unique to this Data Object.
   string id = 1;
 
   // OPTIONAL
-  // The name of the object
-  string name = 2;
+  // A string that can be optionally used to name a file.
+  string file_name = 2;
```
david4096: Remove `file` from variable name.

proto/data_objects.proto
```diff
@@ -1,65 +1,65 @@
...
 
   // REQUIRED
-  // The size, in bytes, of the object
-  int64 size = 3;
+  // The computed file size in bytes.
+  int64 file_size = 3;
```
david4096: Remove `file` from variable name.

proto/data_objects.proto
```diff
@@ -69,30 +69,71 @@ message DataObject {
...
+  string description = 7;
+
+  // OPTIONAL
+  // A list of strings that can be used to identify this Data Bundle.
+  repeated string aliases = 8;
 
   // OPTIONAL
   // A set of key-value pairs that represent system metadata about the object.
-  map<string, string> system_metadata = 12;
+  map<string, google.protobuf.ListValue> system_metadata = 9;
```
david4096: Clarify comment to make clear these are only data bundle level metadata items.

david4096:
Suggested to change the map to Listvalue to a struct to better capture the variety of metadata.

proto/data_objects_service.proto
```diff
@@ -1,195 +1,286 @@
...
+  // A human readable description of the contents of the Data Object.
+  string description = 6;
+
+  // OPTIONAL
+  // A list of strings that can be used to identify this Data Object.
+  repeated string aliases = 7;
+
+  // OPTIONAL
+  // A set of key-value pairs that represent system metadata about the object.
+  map<string, google.protobuf.ListValue> system_metadata = 8;
```
david4096: Remove.

proto/data_objects_service.proto
```diff
@@ -1,195 +1,286 @@
...
 }
 
-message GetDataBundleObjectRequest {
-  string id = 1;
+// The Data Object one would like to index. One must provide any aliases
+// and URLs to this file when sending the CreateDataObjectRequest. It is up
+// to implementations to validate that the Data Object is available from
+// the provided URLs.
+message CreateDataObjectRequest {
+  // OPTIONAL
```
david4096: Add the ability to set the ID.

proto/data_objects_service.proto
```diff
@@ -1,195 +1,286 @@
...
-  string id = 1;
+message GetDataObjectResponse {
+  // REQUIRED
+  // The Data Object that coincides with a specific GetDataObjectRequest.
+  DataObject data_object = 1;
 }
 
-message DataBundleStatusResponse {
+message UpdateDataObjectRequest {
   // REQUIRED
```
david4096: Describe the update semantics using a test case.

proto/data_objects_service.proto
```diff
@@ -1,195 +1,286 @@
...
-  string id = 1;
+  // The identifier of the Data Object to be updated.
+  string data_object_id = 1;
+
+  // OPTIONAL
+  // A string that can be optionally used to name a file.
+  string file_name = 2;
```
david4096: Remove file from variable name.

proto/data_objects.proto
```diff
@@ -69,30 +69,71 @@ message DataObject {
...
+  repeated URL urls = 9;
 
   // OPTIONAL
-  // User controlled brief descriptions of object
+  // A human readable description of the contents of the Data Object.
   string description = 10;
 
   // OPTIONAL
-  // Identifiers of other dataobjects
+  // A list of strings that can be used to identify this Data Object.
```
david4096: Add an example of using aliases to describe hierarchy in S3.

bwalsh:
+1 - pending:
* struct for user & system metadata
* additional comments on aliases. 
* remove file_
* full data object on Create()

david4096:
I've now included the demonstration notebook in the examples directory! https://github.com/david4096/data-object-schemas/blob/341b10cdf3429102276bb83ac63edf5d42f19432/python/examples/gdc_notebook.ipynb

python/ga4gh/dos/client.py
```diff
@@ -0,0 +1,31 @@
+# Simple client usage via bravo
+
+import os
+
+from bravado.client import SwaggerClient
+# from bravado.swagger_model import load_file
+
+SWAGGER_FILENAME = 'data_objects_service.swagger.json'
+current_directory = os.path.dirname(os.path.realpath(__file__))
+SWAGGER_PATH = os.path.join(current_directory, SWAGGER_FILENAME)
```
bwalsh: When the user installs the package, will the swagger file be included?   It would be useful.

bwalsh: Given a dict:
```
{'updated': '2017-10-23T16:20:45+00:00', 'created': '2017-10-23T16:20:45+00:00', 'urls': [{'url': 's3://test/my-bucket/passport+photo.JPG', 'system_metadata': {'bucket_name': 'my-bucket', 'StorageClass': 'STANDARD', 'event_type': 'ObjectCreated:Put'}, 'user_metadata': {'foo': 'bar'}}], 'file_size': 341005, 'checksums': [{'checksum': 'd3b3a66c7235c6b09a55a626861f5f91', 'type': 'md5'}], 'id': 'passport+photo.JPG'}
```
After unmarshal(), there seems to be an extra 'fields' attribute in metadata
```
ga4ghDataObject(aliases=None, checksums=[{'checksum': 'd3b3a66c7235c6b09a55a626861f5f91', 'type': 'md5'}], created=datetime.datetime(2017, 10, 23, 16, 20, 45, tzinfo=tzutc()), description=None, id='passport+photo.JPG', mime_type=None, name=None, size=None, updated=datetime.datetime(2017, 10, 23, 16, 20, 45, tzinfo=tzutc()), urls=[{'url': 's3://test/my-bucket/passport+photo.JPG', 'system_metadata': {u'fields': None, 'bucket_name': 'my-bucket', 'StorageClass': 'STANDARD', 'event_type': 'ObjectCreated:Put'}, 'user_metadata': {u'fields': None, 'foo': 'bar'}}], version=None)
```

david4096:
Hi @bwalsh! I think we've worked out most of the basic data representation issues. I'll be using this work as a basis to demonstrate synthesizing the HCA and GDC data in the coming weeks.

We're using a fork of bravado to support the proto-go-swagger pipeline. This means to install this repo from pip you'll need to use the flag `--process-dependency-links` so that it will properly install the bravado fork. I'm certainly open to other solutions to improving the installation process!

`pip install git+git://github.com/david4096/data-object-schemas@dos-minimal2 --process-dependency-links`

Thanks so much for all your input so far @bwalsh! I'd like to hear from any of @briandoconnor @geoffjentry @philloooo and we can continue to push ahead on schema alterations and tooling in a future PR!

python/ga4gh/dos/client.py
david4096: Yes, the swagger is currently included https://github.com/ga4gh/data-object-schemas/pull/24/files#diff-2eeaed663bd0d25b7e608891384b7298R6 , setting this up to work with release/packaging will be forthcoming.

PR #88, opened 2018-06-19
Packaging improvements

natanlao:


setup.py
```diff
@@ -21,31 +11,6 @@
 with open("README.md") as readmeFile:
     long_description = readmeFile.read()
 
-install_requires = []
-with open("python/requirements.txt") as requirementsFile:
```
david4096: Although I'm ok with removing this convenience as it gives better confidence about the content of the package without using out-of-band info, I might argue it reduces install complexity since we declare the requirements in a single place.

Since this project has very few requirements we might only maintain them in the setup.py, requirements.txt being unecessary?

natanlao: I moved `python/dev-requirements.txt` to `requirements.txt` and inlined all of `python/requirements.txt` in `setup.py` - I don't think that we should add the dev requirements to `setup.py` since those dependencies aren't necessary to someone not developing and just installing off PyPI. (That said, I think it could be prudent to rename `requirements.txt` to `dev-requirements.txt` to make the intent of the file more clear)

Issue #90, opened 2018-06-21
Swagger UI points at inactive service


david4096:
The current github landing pages does not point at a standing service, so "Try It Out!" in the swagger-ui will fail. To close this, point to an active resolver so the github pages swagger UI presents a useful DOS.

natanlao:
#95 disables the "try it out" feature until a resolver is added.

cricketsloan:
➤ David Steinberg commented:

I think we can point to dos-azul-lambda to close this:

[https://dos.commons.ucsc-cgp.org](https://dos.commons.ucsc-cgp.org)

natanlao:
I think that would be a good idea once we have full feature parity in dos-azul-lambda

PR #95, opened 2018-06-22
Update Swagger UI

natanlao:


index.html
```diff
@@ -1,95 +1,75 @@
...
+        }
+      }
     }
 
-    body {
-      margin:0;
-      background: #fafafa;
+      // Build a system
+      const ui = SwaggerUIBundle({
+        url: "https://raw.githubusercontent.com/ga4gh/data-object-service-schemas/master/openapi/data_object_service.swagger.yaml",
```
david4096: Smart :+1:

david4096:
Thanks, this will reduce the needed maintenance on the swagger-ui for a while. We'll come back to this when we have a service we want to use for #90 .

PR #101, opened 2018-06-25
Test Swagger schema validity (closes #100)

### LINKED ISSUE #100: Add Swagger schema validation test

natanlao:


python/test/test_package.py
```diff
@@ -14,7 +20,17 @@ def get_spec_version(cls, filename):
...
-            assert __version__ == self.get_spec_version(os.path.join(spec_dir, f))
+        for f in os.listdir(self.spec_dir):
+            assert __version__ == self.get_spec_version(os.path.join(self.spec_dir, f))
+
+    def test_schema_validity(self):
+        """Check to make sure that the Swagger schema specification is
+        valid by running it through the online validator. If the schema
+        is valid, the online validator will return 200 OK and an empty
+        JSON response."""
+        path = os.path.join(self.spec_dir, 'data_object_service.swagger.yaml')
```
david4096: Thank you, this surface is currently embarrassingly untested and these tests will improve the experience of managing schema changes (since you can run `nostests test/test_package`).

I think it's worth seeing if you can use [swagger-spec-validator](https://pypi.org/project/swagger-spec-validator/) instead of hitting a service. We're implicitly depending on it through bravado:

https://github.com/Yelp/bravado-core/blob/master/setup.py#L19

PR #105, opened 2018-06-26
Fix OpenAPI packaging issue (closes #103)

### LINKED ISSUE #103: OpenAPI schemas packaging problems
The current release has a problem wherein it does not recognize the location of the installed schemas. To replicate `pip install ga4gh-dos-schemas` and then note the error when running `ga4gh_dos_server`.

```
    return io.open(str(self), mode, buffering, encoding, errors, newline)
IOError: [Errno 2] No such file or directory: '/usr/local/lib/python2.7/dist-packages/ga4gh/dos/data_object_service.swagger.yaml'
```

A quick patch is to move the schemas to that directory, this will allow you to start the demo server as normal.

```
wget https://raw.githubusercontent.com/ga4gh/data-object-service-schemas/master/openapi/data_object_service.swagger.yaml
sudo cp /usr/local/lib/python2.7/dist-packages/ga4gh/dos/.
ga4gh_dos_server
```

natanlao:
There were two problems:
1. `data_object_service.swagger.yaml` was not properly being packaged. This happened as a result of some of the intricacies of Python packaging: contrary to intuition, the `use_package_data` flag doesn't really enable the inclusion of package data (and in some cases - such as this one - can inhibit it from working altogether). Additionally, files specified with `package_data` [are included in binary distributions but not source distributions](https://stackoverflow.com/questions/7522250/how-to-include-package-data-with-setuptools-distribute/14159430#14159430) (similarly, it also appears that files specified with `MANIFEST.in` are included in source distributions but not binary distributions). 
2. Even if the file were properly packaged, it would not be accessible with `zip_safe=True`.

This PR fixes the above issue and configures Travis to install the package in a process more closely aligned with an install from PyPI to hopefully catch something like this in the future.

The yaml should now be packaged correctly:
```
$ python setup.py sdist | grep yaml
copying openapi/data_object_service.openapi.yaml -> ga4gh_dos_schemas-0.3.0/openapi
copying openapi/data_object_service.smartapi.yaml -> ga4gh_dos_schemas-0.3.0/openapi
copying openapi/data_object_service.swagger.yaml -> ga4gh_dos_schemas-0.3.0/openapi
copying python/ga4gh/dos/data_object_service.swagger.yaml -> ga4gh_dos_schemas-0.3.0/python/ga4gh/dos
```

This PR closes #103.


david4096:
Thanks, tested using: `pip install git+git://github.com/natanlao/data-obct-service-schemas@issues/103-openapi-packaging` and installing locally using `pip install .`.

PR #111, opened 2018-06-28
Push py3.6 distribution only (closes #110)

### LINKED ISSUE #110: Travis release "fails" when testing two Python versions
Travis builds the project for Python 2.6, sees the build test successfully, then pushes it to PyPI. It then repeats the cycle for Python 3.6 but then "fails" the build because it can't push a source distribution with the same name to PyPI again.

natanlao:
This project is compatible with both Python 2.7 and 3.6
so we don't need to manage separate distributions for each
version. (https://stackoverflow.com/a/46387436)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,8 @@ deploy:
   provider: pypi
   on:
     tags: true
+    branch: master
+    python: '3.6'
   user: david4096
   password:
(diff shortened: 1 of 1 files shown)


david4096:
Righteous, we'll see how it goes on next tagged release!

Issue #118, opened 2018-06-29
Make data bundles checksum optional


david4096:
It is currently required and should not be.

PR #128, opened 2018-07-11
Use 'required' tag instead of comments in schema

### LINKED ISSUE #82: Use JSON schemas required
The required element is now available since we migrated from protobuf, this allows us to be specific in the schema about which elements are required to ease validation.

To close this, edit the schemas to make require all elements tagged required in the comments and demonstrate some validation tests.

### LINKED ISSUE #112: checksum in the GET /dataobject and GET /databundle endpoints
In the GET /dataobject and GET /databundle endpoints checksum is a required field. Is the intended use of this parameter to filter dataobjects and databundles by their checksum values? If not then what is the intended use?

natanlao:
Closes #82. Closes #112.

Given #118 and #127 (i.e. all parameters to `/dataobjects` and `/databundles` are supposed to be optional) is it enough that the endpoints are already implicitly tested in the test suite? https://github.com/ga4gh/data-object-service-schemas/blob/8dc870a10320cf2c03d7cd7d9b4735688f87fd9f/python/test/test_server.py#L295

### DIFF
--- a/python/test/test_server.py
+++ b/python/test/test_server.py
@@ -9,6 +9,7 @@ import unittest
 # setup connection, models and security
 from bravado.requests_client import RequestsClient
 from bravado.exception import HTTPNotFound
+import jsonschema
 
 from ga4gh.dos.client import Client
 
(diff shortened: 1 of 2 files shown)


natanlao:
Tests are failing because of #118 and #127. Since we don't want to break master, and they're both one-line fixes that are related to this ticket, I'll add their fixes to this PR.

natanlao:
This also closes #112.

david4096:
Yes, I expect that this will break some downstream code since we are now using schematized validation. These are good things to break as we strengthen the quality of the schematization.

natanlao:
I misinterpreted the docs - I'll revise this PR.

david4096:
Would you also please add a test that shows the server giving an expected error response when a required field isn't provided? This will be be addressed more fully in https://github.com/ga4gh/data-object-service-schemas/issues/133

natanlao:
Ah sorry, I accidentally pushed to my fork instead of this repo. Here it is.

python/test/test_server.py
```diff
@@ -622,3 +659,20 @@ def test_no_find(self):
...
             self.assertEqual(e.status_code, 404)
+
+    def test_schema_required(self):
+        """
+        Tests that the server properly rejects a request
+        missing a parameter that is marked as required.
+        """
+        CreateDataObjectRequest = self._models.get_model('CreateDataObjectRequest')
+        DataObject = self._models.get_model('CreateDataObjectRequest')
+        # Missing the `id` parameter
```
david4096: Thanks for this!

PR #149, opened 2018-08-01
Improvements to /service-info endpoint

natanlao:


natanlao:
I've manually verified that the generated SmartAPI schema is valid. Thanks for the review @achave11!

PR #151, opened 2018-08-02
Fix schema validation warnings

natanlao:
I can also confirm that the generated SmartAPI schema is reported as valid.


natanlao:
Thanks @achave11!

PR #158, opened 2018-08-21
Add utility that can generate a Swagger schema from a chalice.Chalice.routes object

natanlao:
This PR is the first of two steps that can close DAL-19 (DataBiosphere/dos-azul-lambda#29). This PR adds a function, `ga4gh.dos.schema.from_chalice_routes` that can generate a valid Swagger / OpenAPI 2.0 schema from a `chalice.Chalice.routes` object (which is just a `collections.defaultdict`).

With this utility, implementing an accurate `/swagger.json` in a DOS lambda would be as straightforward as:

    @app.route('/swagger.json')
    def swagger():
        return ga4gh.dos.schema.from_chalice_routes(app.routes)

With this solution, any updates in the functionality of a DOS lambda would be reflected in the exposed `/swagger.json` with little to no manual curation required. Furthermore, this provides the advantage of formally specifying the version of DOS that is implemented by the service by pulling the schema from this package (as opposed to the current solution which returns the latest version of the schema on master).


python/ga4gh/dos/schema.py
```diff
@@ -0,0 +1,56 @@
...
+    # Next, remove from the generated schema paths that are not defined
+    # in :obj:`routes`.
+    schema['paths'] = {k: v for k, v in schema['paths'].items()
+                       if k in routes.keys()}
+
+    # Loop over the remaining paths in the generated schema and remove
+    # methods that are not listed in :obj:`routes`.
+    for path in schema['paths'].keys():
+        schema['paths'][path] = {k: v for k, v in schema['paths'][path].items()
+                                 if k in [r.lower() for r in routes[path]]}
```
david4096: Very nicely done!

python/ga4gh/dos/schema.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+    Given a :obj:`chalice.Chalice.routes` objects, computes the proper
+    subset of the Data Object Service schema and presents it as an
+    OpenAPI 2.0 JSON schema.
+    :rvtype: dict
+    :param chalice.Chalice.routes routes:
+    :param str base_path: the base path of the endpoints listed in `routes`.
+                          This is only necessary if a base path is manually
+                          prepended to each endpoint your service exposes,
+                          e.g. ``@app.route('/ga4gh/dos/v1/dataobjects')``.
```
david4096: This is really great work. It allows one to say, "I made a service with some DOS endpoints" and then generate a swagger doc just for those endpoints that we have documented. This means that implementations in chalice that follow the schema's routes can have documentation generated for them automatically.

PR #160, opened 2018-08-29
Add standardized compliance test suite for DOS implementations

natanlao:
Related to [TLP-492](https://ucsc-cgl.atlassian.net/browse/TLP-492).

This PR implements a standardized compliance test suite that can be used to verify a DOS implementation's compliance with the schema. Specifically, it does the following:
- Implements the test suite present in `dos-azul-lambda` as `ga4gh.dos.test.compliance.AbstractComplianceTest`
- Adds tools to `AbstractComplianceTest` to quickly select which tests should be run
- Refactors `AbstractComplianceTest` to make using the test suite in a generic implementation easier
- Adds documentation for the above in the 'Implementation tools' section of ReadTheDocs
- Correct some deficiencies in the existing tests and in `ga4gh.dos.server`

The rationale for this PR is explained fully in TLP-492. In short, the reasoning is that since all of our lambdas implement the same service, a change to the test suite for one of the lambdas would be valuable in the test suites of the others. Having a single, standard test suite not only reduces duplication of work, but also enables other persons who might be implementing their own data object service to quickly and easily determine the compliance of their installation.

The following remains to be done:
- [ ] Add full schema coverage to compliance test suite. Currently, the test suite only covers the features that `dos-azul-lambda` implements.
- [ ] Percolate this change to the lambdas.
- [ ] Add this to the `ga4gh.dos` test suite. (Tests are passing... for the most part. There appears to be some intermittent bug related to setting up and tearing down the test server.)


david4096:
Really quite beautiful. The usage of decorators to specify operation ID requirements is elegant and easy to understand.

This work, as you've stated, helps improve adoption by allowing other DOS implementations to quickly create useful tests that are generic and extensible. I am happy to see this move forward. The comments are extremely thorough and should be enough to get a developer started. Since the compliance tests inherit from unit test cases, they will be familiar to most programmers.

Regarding your third bullet, intermittent test failures in server, I believe this is due to how the server is started and shut down and should be addressed in a separate PR.

david4096:
Nice! The compliance suite tests now run with the rest of nosetests.

image

PR #162, opened 2018-08-30
Roadmap for the next 6 months

david4096:
tl;dr DOS schemas are stable and with two small, non-breaking additions will expand its possible use cases. Supporting software and a suite of compliance tests help to ease adoption. A stand-alone application based around the Data Object Service is released.

https://github.com/ga4gh/data-object-service-schemas/issues/139

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -1,91 +1,108 @@
 Roadmap
 -------
 
+The Data Object Service has been implemented in multiple workflow execution engines 
+and has helped to allow cloud agnostic data access. A small network of DOS servers 
+is available from a resolver at dataguids.org.
+
+Software for presenting Human Cell Atlas, NCI GDC, GTEx, and TOPMed data are all made 
+available. Making it easier to discover, interact with, access, and publish new data 
+using supporting software, documentation, and possible schema changes make up the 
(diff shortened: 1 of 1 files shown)


ROADMAP.md
```diff
@@ -1,91 +1,108 @@
...
+bulk of this roadmap.
 
-## Metrics of success
+## Goal
 
-What data are available by DOS? GDC? TOPMed? HCA? How many files/bundles/PB?
+### Short term - 6 months
 
-How many workflows have been converted to use DOS downloader?
+Release version 1 and improve ease of use and adoption through supporting software.
```
geoffjentry: @david4096 Considering the cloud workstream hasn't even started discussing and/or developing DOS, doesn't this seem like an aggressive timeline?

david4096: Although it's hardly the case the workstream hasn't discussed DOS, this release candidate would be the one we would move forward with the approval process on, I presume. @briandoconnor  ?

geoffjentry:
There has been no official work on DOS, all official focus since the reboot has been on WES. We’ll need to loop in the driver projects in a serious way prior to pushing towards 1.0

PR #164, opened 2018-09-06
Add build stages (closes #161)

### LINKED ISSUE #161: Travis can upload package if build fails with Python 2.7

natanlao:
With build stages, the build will fail if one job in a stage fails.
So, if in the testing stage, the Python 2.7 job fails, the build
will stop and not run the deploy, even if the Python 3.6 job completes
successfully.


natanlao:
@jessebrennan

.travis.yml
```diff
@@ -1,21 +1,53 @@
...
+branches:
+  only:
+    - master
+    - /^\d+\.\d+(\.\d+)?(-\S*)?$/
+
+stages:
+  - linting
+  - test
+  - name: deploy
+    if: type != pull_request
```
jessebrennan: Does this need to be here?

natanlao:
Thanks for the review @jessebrennan!

PR #166, opened 2018-09-06
Load schema locally by default (closes #114)

### LINKED ISSUE #114: Load open API description locally for client
The client currently requires that a service present an OpenAPI description in order to set up its connection. That shouldn't be necessary. To close this change the client to load from file. It should be accompanied by work on reasoning between versions.

natanlao:
Refactor :class:`ga4gh.dos.client.Client` to allow loading the schema
locally for implementations that may not present their own ``swagger.json``.
This is done by introducing a ``local`` parameter:

    from ga4gh.dos.client import Client
    client = Client(url='...', local=True)

This feature is not enabled by default, so the previous functionality is
preserved::

    client = Client(url='...')

I'm submitting this from out of sprint because it would make closing DataBiosphere/dos-azul-lambda#42 easier.

### DIFF
--- a/python/ga4gh/dos/client.py
+++ b/python/ga4gh/dos/client.py
@@ -10,11 +10,18 @@ following the OpenAPI schema.
 It currently assumes that the service also hosts the swagger.json, in a style
 similar to the demonstration server, :mod:`ga4gh.dos.server`.
 """
+try:  # for python3 compat
+    import urlparse
+except ImportError:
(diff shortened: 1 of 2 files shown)


natanlao:
I'm not seeing the checkmark here but tests are passing: https://travis-ci.org/ga4gh/data-object-service-schemas/builds/426837653

david4096:
Thanks!

Issue #47, opened 2018-02-16
Authorization metadata per URL


david4096:
Some DOS installations present private and controlled access data. To make it easier for clients to find out the proper authentication credentials to use, add a `ServiceInfo` that offers a URL for logging in.

zflamig:
Need to consider that the DOS installation which knows about the data may not be the same one which can access the data doing it in one shot with sending the DOS server your auth info may not be the best way.

How will we know which protocols the auth URL speaks? There may need to be a standard around that too or a list of options... consider FTP with anonymous access or quasi anonymous (email registered with any password) or HTTP with basic auth only?

david4096:
In order to avoid enumerating all the possible schemes different URLs might enforce, my hope was to offer this as a simple matter of service with some helpful URLs. You're right that the same service might host several URLs that all offer different authentication mechanisms.

To that end, it seems like a list of pairs of protocols and URLs might offer a minimum workable solution.

For example, imagine that we host the Thousand Genomes data via DOS. We list both S3 URLs and ftp locations for each item. We want to convey to the client enough information so that can properly form a request that includes their credentials, or a way to begin to gather credentials.

To currently satisfy this use case, note that the `url` field of a Data Object can include metadata. Without other guidance, one could define keys in the `user_metadata` that could instruct a client to set anonymous FTP details, or that the S3 data was publicly downloadable without requester pays.

Since we expect these metadata to be shared across many items in this repository, we could add these tuples to the `ServiceInfo` endpoint to make its use more clear and to help constrain the useful keys. If there are multiple authentication schemes for one protocol type (http basic auth, oauth...) for Objects behind a service, this could get complicated.

We may find in practice that a limited useful set of terms will be used, for example, doing this for [fence](https://github.com/uc-cdis/fence) one might want to include the "authorize" and "token" endpoints. This would make it easier for a "fence aware downloader" to work with a DOS.

Perhaps instead of trying to add an endpoint, we could constrain the keys used for describing the URL metadata. So, add a new key to the URL field of a data object (which is already an object that can include `system_metadata` and `user_metadata`) that is something like `authorization_metadata` and be instructive about useful keys? https://github.com/ga4gh/data-object-service-schemas/issues/25

david4096:
A target for interoperation for this service's response would be in assisting clients in generating a `keychain.json` of the bdbags downloader. https://github.com/fair-research/bdbag/blob/master/doc/config.md#keychain.json

[...]

PR #167, opened 2018-09-07
Add auth metadata key to address #47

david4096:
This is an attempt to address #47 by accepting a suggestion provided by @mfiume.

This will allow services to provide optional metadata fields for each URL under the `authorization_metadata` key. [As mentioned in](https://github.com/ga4gh/data-object-service-schemas/issues/47#issuecomment-402524882) #47 , we need to determine in practice which keys will be necessary, however, adding this key will provide a single place for those metadata to be found.

The problem being addressed is to allow downstream software that uses Data Objects to be able to delegate auth concerns as necessary. For example, once a client has resolved the URL and auth metadata for a Data Object, a downstream WES would like to be able to request authorization for accessing that object from the underlying object store as necessary.

### DIFF
--- a/openapi/data_object_service.swagger.yaml
+++ b/openapi/data_object_service.swagger.yaml
@@ -562,6 +562,20 @@ definitions:
             A set of key-value pairs that represent metadata provided by the uploader.
   AuthorizationMetadata:
     type: object
+    properties:
+        provider:
+          type: string
+          description: |-
+            The auth provider, e.g. auth0, google, etc.
+        auth_type:
+          type: string
+          description: |-
+            The auth standard being used to make data available. For example, 'OAuth2.0'.
+        auth_url:
+          type: string
+          description: |- 
+            The URL where the auth service is located, for example, a URL to get an OAuth
+            token.
     additionalProperties: true
     description: |-
             OPTIONAL
--- a/openapi/data_object_service.swagger.yaml
+++ b/openapi/data_object_service.swagger.yaml
@@ -560,6 +560,13 @@ definitions:
     description: |-
             OPTIONAL
             A set of key-value pairs that represent metadata provided by the uploader.
+  AuthorizationMetadata:
+    type: object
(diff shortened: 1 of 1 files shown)


openapi/data_object_service.swagger.yaml
```diff
@@ -560,6 +560,13 @@ definitions:
     description: |-
             OPTIONAL
             A set of key-value pairs that represent metadata provided by the uploader.
+  AuthorizationMetadata:
+    type: object
+    additionalProperties: true
+    description: |-
+            OPTIONAL
+            A set of key-value pairs that represent sufficient metadata to be granted
```
briandoconnor: Don't we need some structure here?  Like if I was building a generic DOS client how would I know what these fields would be ahead of time?  And if I wanted to support a particular service but everyone describes that service with different terms how will that work?

mfiume:
Thanks so much for this @david4096. I agree with @briandoconnor that it would be very useful to standardize a minimal set of keys that would be sufficient for a consumer of this data model to perform an authorized transfer of the object without having prior knowledge of the source.

zflamig:
Yeah. I agree with the others we should define at least a minimal set of keys and values to start. Something for just OIDC flows?

david4096:
Great, I'll change it to use the keys you proposed @mfiume . If anyone has any suggestions for them, I'll include them!

zflamig:
Do we need the provider part of this? I'm wondering if it will just add confusion... for example it doesn't really differentiate between the tool the entity is using and the IDP?

david4096:
@mfiume do you have a strong opinion on the matter?

david4096:
I removed the provider key and mentioned it as something to consider for additional properties.

briandoconnor:
I think this is going to need more work, see my comments. But to move this forward since it's a step in the right direction I'm merging.  We can tweak based on community feedback in the transition to DRS by our drivers.

PR #171, opened 2018-09-14
Make checksum in ListDataObj optional (fixes #170)

### LINKED ISSUE #170: `checksum` parameter in ListDataObjectsRequest should be optional
This was supposed to be done in #127 - I forgot to fix the parameter in the ListDataObjectsRequest model.

natanlao:


david4096:
+1

PR #183, opened 2018-10-11
Add compliance testing notebook (closes #176)

### LINKED ISSUE #176: Add notebook demonstrating compliance tests
> follow up compliance work with a notebook that runs tests on all the different DOS' we know about

natanlao:
@david4096, is this about what you're looking for? Is there anything I should add or clarify? I'm not sure making the code directly runnable would be valuable since there's a good amount of boilerplate needed to run unittest test cases

### DIFF
--- a/python/examples/compliance_testing.ipynb
+++ b/python/examples/compliance_testing.ipynb
@@ -1,0 +1,289 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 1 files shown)


david4096:
This is great @natanlao ! If it's not too much trouble it would be interesting to run the tests against a source or two to show the results? You might also add a brief conclusion saying something along the lines of, "this is part of larger effort to automate the testing and compliance of DOS implementations."

natanlao:
Pushed a new update - now, the notebook should run all the code nicely.

PR #185, opened 2018-10-25
Update schema for SmartAPI upload

natanlao:
The `x-role` key is needed for SmartAPI compliance


natanlao:
Schemas are verified for both SmartAPI and Swagger

PR #209, opened 2019-01-21
Doc build improvements (first of multiple)

natanlao:
Small improvements to Swagger UI distribution - the entire package is 8.8M. By replacing this with a page that frames Swagger UI, we can reduce the disk usage overhead to 8.0K with no real functional regression.


natanlao:
Actually, looking at this again, I am not happy with this. I'm going to close this and open another, cleaner PR that addresses this issue in a better way.

Issue #203, opened 2019-01-17
Improve doc build system


briandoconnor:
James E. gave some feedback that we can work on in a new PR:

.travis.yml
@@ -35,6 +35,25 @@ jobs:
        # Travis will install requirements.txt by default
        - flake8 --select=E121,E123,E126,E226,E24,E704,W503,W504 --ignore=E501 app.py tests

     # the build process from James Eddy adapted from the WES repo
  @jaeddy
jaeddy on Dec 14, 2018  Member
Just a note: the steps/stages described under jobs should be able to replace the python and stages sections above. Might be a task for a separate PR.

briandoconnor:
gulpfile.js
@@ -0,0 +1,50 @@
var gulp = require('gulp');
  @jaeddy
jaeddy on Dec 14, 2018  Member
I'm about 95% sure that this file isn't needed anymore. It's leftover from when I was using a separate node job in Travis to build the Swagger UI.

natanlao:
It looks like `ga4gh.github.io/data-repository-service-schemas/swagger-ui/` etc. is broken

jaeddy:
I've come across that before with the WES repo — pretty sure it's a bug in the logic of either the `fetchpages.sh` or `stagepages.sh` script (in particular, the `if [ "$TRAVIS_BRANCH" <something> ]` statements). I'll take a closer look and try to remind myself what's happening here...

jaeddy:
+1 

Awesome — that sounds great, @natanlao!

Something else that could be useful when addressing this bug is adding some print/echo calls for Travis environment variables (e.g., `"$TRAVIS_BRANCH"`, `"$TRAVIS_PULL_REQUEST"`, `"$TRAVIS_PULL_REQUEST_BRANCH"`) such that they're visible in the build logs. Maybe it's just me, but this currently feels a bit like a black box, which makes it harder to predict behavior.

natanlao:
I think the only thing needed to close this ticket is to add some test to ensure that the build system is working as expected.

briandoconnor:
Related to #230 .  Stale issue, can we close?

jb-adams:
Suggest closing this issue, as #359 updates the docs build process to use [gh-openapi-docs](https://github.com/ga4gh/gh-openapi-docs)

@briandoconnor @dglazer

PR #210, opened 2019-01-21
Doc build improvements (second of multiple)

natanlao:
This is a separate PR since some changes need to be made against `gh-pages` and others against `develop`.

"Improvements" is vague, but I think it's okay this time around given the hackathon setting and since #203 says "Improve doc build system"


natanlao:
Note that the system won't be fully functional until develop is merged into master, as the current master configuration won't build any branch except for master and tagged commits

scripts/stagepages.sh
```diff
@@ -5,16 +5,16 @@ set -v
...
-  branchpath="preview/$branch"
-  mkdir -p "$branchpath/docs"
-  cp docs/html5/index.html "$branchpath/docs/"
-  cp docs/pdf/index.pdf "$branchpath/docs/"
-  cp openapi/data_repository_service.swagger.yaml "$branchpath/swagger.yaml"
-  cp -R web_deploy/* "$branchpath/"
+    branch=$(echo "$TRAVIS_BRANCH" | awk '{print tolower($0)}')
+    branchpath="preview/$branch"
+    mkdir -p "$branchpath/docs"
+    cp docs/html5/index.html "$branchpath/docs/"
```
jaeddy: At some point, we could consider removing the `docs/html5/` folder from any non-`gh-pages` branches. The `index.html` artifact isn't especially useful as a downloadable / version controlled file.

Not critical for now.

Issue #213, opened 2019-01-23
Object metadata and download methods 


sarpera:
# Background

Following the discussion we had at the [GA4GH hackathon in January](https://docs.google.com/document/d/1cFcv5wdsfXI8SsEKtJcPqKTrxg5sCexLiyPy9kC-I5E/edit?usp=sharing)
we would like to propose to have a method to get the metadata of an object, 
and then have an additional method which will provide the download of the 
object.

The rationale to have two methods instead of one, is due to the necessity
to sign the object using the authorisation token provider (right now this
is based on the OIDC specs), which is expensive computationally to do.
More over, with the presence of regions and provider, a DRS client will 
be able to decide which provider and which region would be best to obtain 
the file, among all the possible URIs.

The format we propose are:
    - ```objects/<id>/``` for getting the object metadata
    - ```objects/<id>/download``` for getting the object bytes

and we propose to pass the authorisation token in the Request Header 
to get access to the object.

This is the flow, from a DRS client point of view:

1) GET `/objects/<id>`

2) GET `/objects/<id>/download` with Request Header `X-DRS-TOKEN: <TOKEN>`

The token is obtained by the client from the DRS server, and it is up to the DRS Server 
implementer to decide how a user will obtain that.

## Object metadata Request

This will return the object metadata:

### HTTP Request

```
GET /objects/<id>

```

### HTTP Response

```json
{
  "object": {
    "id": "string",
    "name": "string",
    "size": "string",
    ...
    "urls": {
      "cloud": [
        {
          "uri": "s3://<foo>/<bar>.bam",
          "region": "us-east-1",
          "provider": "aws"
        },
        {
          "uri": "gs://<foo>/<bar>.bam",
          "region": "us-west1",
          "provider": "google"
        }
      ],
      "ftp": [
        {
          "uri": "ftp://foo.com/bar.bam"
        }
      ],
      "drs": [
        {
          "uri": "drs://foo.com/objects/<id>"
        }
      ]
    },
    "aliases": [
      "doi://123/abcd"
    ]
  }
}
```

The client will be able to pick one of the `cloud` uri and request 
the download uri, passing the token

### HTTP REQUEST

```
GET /objects/<id>/download?type="cloud"&uri="gs://<foo>/<bar>.bam" 
    
    Request Header: 
    X-DRS-TOKEN: <TOKEN>

```

### HTTP Response

The return value is a URI where a GET request will give you the bytes:

```json
{
[...]

PR #236, opened 2019-02-28
Issue #213: Access methods and access path for an object

sarpera:
Issue #213 - Access methods and access path for an object

**Summary:**
- Replaced `urls` property with `access_methods` in an object response
- An access method can have: `type*`, `access_url`, `access_id`, `region`
- Added `/access/{access_id}` path
- Added `security` definition (optional) API-wide
  - `apiKey` type, in the request header (`Authorization: <token_string>` or `Authorization: Bearer <token_string>`)

**Notes:**
- Items in `access_methods` will be strongly typed per enumerated set of access methods when we move to OpenAPI v3.*. 
  - swagger 2.0 doesn't support `anyOf`, `oneOf`, `allOf`
- `securityDefinitions` are defined much cleaner in OpenAPI v3.*, for now the value for `Authorization` request header is assumed to be any string and optional to provide.

**Feedback needed on:**
- What's the general take on enumerated access method `type`s? 
  - `type` specific properties e.g `region` for cloud access methods
  - extensibility vs strongly typed access methods for clarity
- Examples for existing auth models from drivers
- We can also add `OAuth2` in `securityDefinitions` based on drivers' input
  - `/service-info` can tell which auth model is supported (`apiKey`, `OAuth2`, all, none etc)
  - requires DRS to define `authorizationUrl` and `tokenUrl` on the endpoint
  - example definition:
```yaml
OAuth2:
  type: oauth2
  flow: accessCode
  authorizationUrl: /oauth/authorize
  tokenUrl: /oauth/token
```

### DIFF
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -344,7 +344,6 @@ definitions:
         - globus
         - aspera
         - gsiftp
-        - nfs
         - local
         description: >-
           Type of the access method.
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -18,6 +18,9 @@ consumes:
   - application/json
 produces:
   - application/json
+security:
+  - {}
+  - authToken: []
 paths:
   '/service-info':
     get:
@@ -109,8 +112,6 @@ paths:
     get:
       summary: Returns a fully resolvable URL that can be used to fetch the actual object bytes.
       operationId: GetAccessURL
-      security:
-        - authToken: []
       responses:
         '200':
           description: The access URL was found successfully.
@@ -172,24 +173,6 @@ definitions:
     description: |-
             OPTIONAL
             A set of key-value pairs that represent metadata provided by the uploader.
-  AuthorizationMetadata:
-    type: object
-    properties:
-        auth_type:
-          type: string
-          description: |-
-            The auth standard being used to make data available. For example, 'OAuth2.0'.
-        auth_url:
-          type: string
-          description: |-
-            The URL where the auth service is located, for example, a URL to get an OAuth
-            token.
-    additionalProperties: true
-    description: |-
-            OPTIONAL
-            A set of key-value pairs that represent sufficient metadata to be granted
-            access to a resource. It may be helpful to provide details about a specific
-            provider, for example.
   Checksum:
     type: object
     required: ['checksum']
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -105,6 +105,59 @@ paths:
       tags:
         - DataRepositoryService
       x-swagger-router-controller: ga4gh.drs.server
+  '/objects/{object_id}/access/{access_id}':
+    get:
+      summary: Returns a fully resolvable URL that can be used to fetch the actual object bytes.
+      operationId: GetAccessURL
+      security:
+        - authToken: []
+      responses:
+        '200':
+          description: The access URL was found successfully.
+          schema:
+            $ref: '#/definitions/GetAccessURLResponse'
+        '400':
+          description: The request is malformed.
+          schema:
+            $ref: '#/definitions/ErrorResponse'
+        '401':
+          description: The request is unauthorized.
+          schema:
+            $ref: '#/definitions/ErrorResponse'
+        '404':
+          description: The requested access URL wasn't found
+          schema:
+            $ref: '#/definitions/ErrorResponse'
+        '403':
+          description: The requester is not authorized to perform this action.
+          schema:
+            $ref: '#/definitions/ErrorResponse'
+        '500':
+          description: An unexpected error occurred.
+          schema:
+            $ref: '#/definitions/ErrorResponse'
+      parameters:
+        - name: object_id
+          in: path
+          required: true
+          type: string
+        - name: access_id
+          in: path
+          required: true
+          type: string
+          description: An 'access_id' from 'access_methods' list of a Data Object
+      tags:
+        - DataRepositoryService
+      x-swagger-router-controller: ga4gh.drs.server
+securityDefinitions:
+  authToken:
+    description: |
+      A valid authorization token must be passed in the 'Authorization' header.
+      Example syntax for using 'Authorization' header :
+        Bearer: token_string
+    type: apiKey
+    name: Authorization
+    in: header
 definitions:
   SystemMetadata:
     type: object
@@ -211,7 +264,7 @@ definitions:
         $ref: '#/definitions/UserMetadata'
   Object:
     type: object
-    required: ['id', 'size', 'created', 'checksums']
+    required: ['id', 'size', 'created', 'checksums', 'access_methods']
     properties:
       id:
         type: string
@@ -252,12 +305,13 @@ definitions:
           $ref: '#/definitions/Checksum'
         description: |-
           The checksum of the Data Object. At least one checksum must be provided.
-      urls:
+      access_methods:
         type: array
+        minItems: 1
         items:
-          $ref: '#/definitions/URL'
+          $ref: '#/definitions/AccessMethod'
         description: |-
-          The list of URLs that can be used to access the Data Object.
+          The list of access methods that can be used to access the Data Object.
       description:
         type: string
         description: |-
@@ -282,20 +336,53 @@ definitions:
     properties:
       object:
         $ref: '#/definitions/Object'
-  URL:
+  GetAccessURLResponse:
     type: object
-    required: ['url']
+    required: ['object']
     properties:
-      url:
+      access_url:
         type: string
-        description: |-
-          A URL that can be used to access the file.
-      system_metadata:
-        $ref: '#/definitions/SystemMetadata'
-      user_metadata:
-        $ref: '#/definitions/UserMetadata'
-      authorization_metadata:
-        $ref: '#/definitions/AuthorizationMetadata'
+        description: A fully resolvable access_url that can be used to fetch the actual object bytes.
+  AccessMethod:
+    type: object
+    required:
+      - type
+    properties:
+      type:
+        type: string
+        enum:
+        - s3
+        - gs
+        - ftp
+        - sftp
+        - http
+        - https
+        - nfs
+        - globus
+        - aspera
+        - gsiftp
+        - nfs
+        - local
+        description: >-
+          Type of the access method.
+      access_url:
+        type: string
+        description: >-
+          A fully resolvable URL string that can be used to fetch the actual object bytes.
+          Note that at least one of access_url and access_id must be provided.
+      access_id:
+        type: string
+        description: >-
+          An arbitrary string to be passed to the /access path to fetch an access_url.
+          This must be unique per object.
+          Note that at least one of access_url and access_id must be provided.
+      region:
+        type: string
+        description: >-
+          OPTIONAL
+          Name of the region in the cloud service provider that the object belongs to.
+        example:
+          us-east-1
   ErrorResponse:
     description:
       An object that can optionally include information about the error.


dglazer:
Thanks @sarpera -- looking good based on a very quick skim; I'll respond with finer-grain comments over the weekend.

openapi/data_repository_service.swagger.yaml
```diff
@@ -105,6 +105,59 @@ paths:
       tags:
         - DataRepositoryService
       x-swagger-router-controller: ga4gh.drs.server
+  '/objects/{object_id}/access/{access_id}':
+    get:
+      summary: Returns a fully resolvable URL that can be used to fetch the actual object bytes.
+      operationId: GetAccessURL
+      security:
```
dglazer: Feels like this PR should either specify security on both methods or neither. I'm okay either way, although I lean slightly towards leaving it out entirely here, since we already know we need better auth support (or at least documentation) in a future PR.

Or we can add exactly the same `security` on both methods. If we do keep it, two questions since I'm not familiar with Swagger `security` syntax:
- does this mean an authToken is **required**? I think we want it to be up to the implementor
- it looks like there's also a way to explicitly say "use OAuth2" -- do we want to support that? (Again optionally.)

openapi/data_repository_service.swagger.yaml
```diff
@@ -105,6 +105,59 @@ paths:
...
+          type: string
+        - name: access_id
+          in: path
+          required: true
+          type: string
+          description: An 'access_id' from 'access_methods' list of a Data Object
+      tags:
+        - DataRepositoryService
+      x-swagger-router-controller: ga4gh.drs.server
+securityDefinitions:
```
dglazer: It looks like `AuthorizationMetadata` (below) is no longer used anywhere -- it should probably be deleted.

openapi/data_repository_service.swagger.yaml
```diff
@@ -282,20 +336,53 @@ definitions:
     properties:
       object:
         $ref: '#/definitions/Object'
-  URL:
+  GetAccessURLResponse:
```
dglazer: Does this still need to be an object? Looks like it's now just a string, which would simplify syntax. (I'm open to others input here, since I can imagine that we'll end up adding some other properties down the road.)

openapi/data_repository_service.swagger.yaml
sarpera: Ideally all of the API methods would optionally have `security` applied. That way the implementors could choose what to expose publicly and privately for a user. Also please note that all existing methods define 401 error responses, assuming they would all have some sort of auth in place by design.

Current PR suggests requiring authToken only for the `/access` path. Swagger allows making it optional, as well as requiring OAuth2 optionally. Something like this:
```yaml
security:
  - {}
  - authToken: []
  - OAuth2: [] 
```
Which means: "No auth OR authToken OR OAuth2"

If the above `security` is applied to the whole API, individual methods can also override the security rules e.g /service-info method could be:
```yaml
security:
  - {}
```

My suggestion is to update the PR with the following in mind:
- `authToken` is optional in `security` definition
- apply `security` API-wide, ie. all existing GET methods
- leave `OAuth2` for further discussion

That way no existing work would get affected.
As long as `/access` path has a `security` rule (i.e current PR), it'll cover the cases for private/controlled data in the cloud.

openapi/data_repository_service.swagger.yaml
sarpera: +1

openapi/data_repository_service.swagger.yaml
sarpera: All responses so far have been an object, so I kept it that way. There might be some cases where redirection may not be possible as a response and/or we might add more fields in the future as you mentioned (expiry date etc).

dglazer: Sounds good -- I'm happy leaving it an object for consistency and future-proofing. Resolving.

openapi/data_repository_service.swagger.yaml
dglazer: Great -- we can resolve once you update the PR.

openapi/data_repository_service.swagger.yaml
dglazer: I like your suggestion ("*apply `security` API-wide, with `authToken` as optional, and `OAuth2` deferred for later*"). Up to you about requiring `service-info` to be public; I kind of like that.

sarpera: Great! On a second thought, making authToken optional can cover both cases for `service-info`; public & private. So I'll just mark it API-wide. Updating PR in a moment.

sarpera:
@dglazer, PR is updated based on the input.

openapi/data_repository_service.swagger.yaml
dglazer: update LGTM; resolving

openapi/data_repository_service.swagger.yaml
dglazer: thanks for removing; resolving

dglazer:
@sarpera , thanks for addressing my comments -- I'm going to approve now. As discussed, there are more changes coming, but I think this PR does a good job of locking in the two methods we discussed at the hackathon, and giving us a good foundation for those future changes.

sarpera:
@dglazer @briandoconnor, updated with more detailed explanations and feedback request from drivers.

susheel:
@sarpera @dglazer @briandoconnor I fully support of this PR and will be able to make use of `access_urls` for public data. We will not be able to use this specification as-is for any protected resource within EMBL-EBI/ELIXIR.

We will be looking forward to the next wave of changes, esp. around the previously discussed `access_url_url` (name TBD) and/or `contact` fields to support protected resources.

dglazer:
@susheel -- great, thank you -- let's get this PR buttoned down, and then talk about next steps to support more of your use cases.
@sarpera -- I don't see the new changes you mention (_"more detailed explanations and feedback request from drivers"_) -- where should I be looking?

sarpera:
@dglazer I edited my first message in this thread.

dglazer:
Got it, thanks -- you edited https://github.com/ga4gh/data-repository-service-schemas/pull/236#issue-257106034 -- I was looking at the PR itself, not this thread.

rishidev:
@ddietterich @delagoya - This is now ready to be merged. Please can you vote +1, 0, or -1 on this PR by 2019-03-08 and provide an explanation if you provide a -1. No vote will be regarded as a 0.

ddietterich:
👍 +1

openapi/data_repository_service.swagger.yaml
```diff
@@ -282,20 +319,53 @@ definitions:
...
+        - gs
+        - ftp
+        - sftp
+        - http
+        - https
+        - nfs
+        - globus
+        - aspera
+        - gsiftp
+        - nfs
```
jorizci: nfs appears twice on the enum.

jorizci:
👍 +1

openapi/data_repository_service.swagger.yaml
sarpera: @jorizci thanks, good catch! Corrected and pushed to PR. FYI @dglazer

openapi/data_repository_service.swagger.yaml
```diff
@@ -282,20 +319,52 @@ definitions:
...
       object:
         $ref: '#/definitions/Object'
-  URL:
+  GetAccessURLResponse:
     type: object
-    required: ['url']
+    required: ['object']
     properties:
-      url:
+      access_url:
```
briandoconnor: So it looks like we can just returned signed URLs?  In the previous response that came back we returned a URL along with system_metadata, user_metadata, and authorization_metadata.  In the NIH Data Commons we needed the ability to provide more context about the URL to allow a user to make a request.  Also, in TOPMed and HCA we need to pass back native URLs to s3 and google storage.  Which will work here.  But there's no way that I can see in the current schema to pass back temporary credentials to access those native URLs.  I think this would come up with other protocols like globus or aspera, the users is going to need more than a URL.  What do you think is the best way to support this?  It feels like this is the part of the response that would need to include that info.

delagoya:
OK, hate to be the voice of dissent here, but exactly why are we encoding a protocol ontology in our spec? Seems like a anti-pattern. If this is a need of the client and/or server implementation, I would propose that it be taken care of at that level, rather than the spec itself.

dglazer:
@delagoya , re "_why are we encoding a protocol ontology in our spec_" -- do you mean by having the list of valid `AccessMethod type` values (s3:, gs:, http:, etc.)? If so, I think the original intent was to have type-specific parameters, so that the spec could say (for example) "region is required for these types, optional for these, and invalid for others". That got watered down because of OpenAPI v2 limitations, but might get reintroduced once we figure out how to switch to OpenAPI v3.

Having said that, even without type-specific parameters, I think there's value in signaling that we expect most DRS servers to implement a subset of a short list of protocols, and there's certainly value in making sure that everyone who supports (e.g.) AWS storage spells it the same way. I could easily get talked into an even shorter list of protocols, that we think cover 80+% of the cases, with extensions beyond that left up to individual server extensions. 

Or am I misunderstanding your concern?

delagoya:
That was my concern.

Even when we get to OpenAPI 3, I would still have misgivings about creating
a protocol ontology within a API spec. Seems wrong somehow. Or it may be
battle scars from previous standards efforts that went whole-hog into
ontology-development-as-code (*cough ... MAGE-ML ... cough*). What about
specialized services/protocols for new or specialized clients? For example,
if Mythical Company X has a smart client that can ask for a specific
protocols based on application need or source type (e.g.
"mcx://mybucket/na12878.cram" would request `htsget` protocol for small
random intervals and stream HTTP for continuous segments).

In any case, if the rest of the group want to use this to signal a limited
set of supported protocols, that is OK with me. (disagree and commit!)

One enhancement if this is accepted: should we be putting into
`/service_info` a  section on `supported_protocols`?

geoffjentry:
I like that `service_info` suggestion

@delagoya for better or for worse (I think a case could be made other way) we seem to have trended towards the side of baking these things in on other APIs and I like striving for consistency. The flip side is that we should all be vigilant in a ["the line must be drawn here. this far and no farther!"](https://www.youtube.com/watch?v=tleSnj4OD0g) manner for the more ridiculous forms of creep

pgrosu:
Oh those good old days of MAGE-ML...what memories...  Anyway, might as well abstract it to a *Unit-of-Work* pattern where the connection and transaction patterns are encapsulated - I'm kidding of course ;)  I'm just waiting for the big day of May 1 when [Terra](https://terra.bio/) will make everything magically work together.

sarpera:
@delagoya I understand the concern here. The intention here is to help the client parse and consume the response programatically, with the least amount of ambiguity.  

If a DRS is serving objects and offering various means of access to bytes via commonly used methods, clients can expect additional information based on the "type" of the access method; e.g you will be interested in the expiry date of a signed URL, region and the provider (egress costs) only for the cloud-based storages (s3, gs etc). Enumerating the "types" enables us to couple those additional information per access method type (well, properly with v3.0).

There is always room to define an access method generically, that would just have `access_url` property and nothing else. You then rely on the client's ability to understand and consume it with no other information.

In your example:
> if Mythical Company X has a smart client that can ask for a specific
protocols based on application need or source type (e.g.
"mcx://mybucket/na12878.cram" would request `htsget` protocol for small
random intervals and stream HTTP for continuous segments).

I would expect DRS to provide me a structured response so that I can consume this information in my ecosystem. Aside for "mcx", my DRS client should tell apart ftp, s3, gs, http programatically and make decisions on the go. I would need to know `htsget` protocol is under the hood and the response should tell me that. 

Note that `type` for an access method in this PR is not equal to `protocol`. The enumeration of the `type`s is intended to tell access methods apart, set expectations for clients, and couple additional information per a certain type if needed.

briandoconnor:
@sarpera what do you think of my request for 301 support?  It's a direct driver project request so I don't want to necessarily ignore it.  Otherwise my second comment about the limitations on returning just access_url can be cleared since I'm happy for this PR to move forward and address that in a future PR as we collect more feedback from drivers.

openapi/data_repository_service.swagger.yaml
```diff
@@ -105,6 +108,57 @@ paths:
...
+            $ref: '#/definitions/ErrorResponse'
+        '404':
+          description: The requested access URL wasn't found
+          schema:
+            $ref: '#/definitions/ErrorResponse'
+        '403':
+          description: The requester is not authorized to perform this action.
+          schema:
+            $ref: '#/definitions/ErrorResponse'
+        '500':
```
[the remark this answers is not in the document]
briandoconnor: See https://github.com/ga4gh/data-repository-service-schemas/issues/238

@sarpera let me know if you just want to push back on this and do it in another PR

rishidev:
Driver Project representatives were emailed on 2019-03-08 for NCI GDC, ENA/EVA/EGA, Genomics England, CanDIG, TopMed, HCA (input provided via Brian) with request for comments and votes by registered 2019-03-18.

sarpera:
@briandoconnor I'll write detailed comments in the respective issues you created, but in short, if we end up incorporating those points they could be easily added separately after this PR goes through I suppose?

delagoya:
@sarpera got it. I am satisfied with this.

dglazer:
Brian, thank you for splitting those off into separate issues -- I agree
with your suggestion and with Sarper that they can easily be addressed in
separate PRs.

dglazer:
Re the list of `type` values for `AccessMethod` -- I think the list in the PR is good for now, and am glad @delagoya and @sarpera are in sync. We might want to refine the list later -- I've opened #240 to discuss.

rdeborj:
Within CanDIG, we will probably use the access path for information purposes only and will be using federated credentials outside DRS.

openapi/data_repository_service.swagger.yaml
sarpera: Answered in [#239](https://github.com/ga4gh/data-repository-service-schemas/issues/239#issuecomment-472534495).

dglazer:
@rdeborj, glad this works for CanDIG. I'd like to understand your use case a little better -- are you saying that your clients would call `GET /objects/{object_id}`, and use the `access_url` returned from there to fetch object bytes, but when doing that fetch would use auth credentials that they obtained via some non-DRS approach? If so, that makes sense to me, and is one of the auth patterns we anticipated. (We'll be beefing up the documentation to make that clearer.)

Or are you saying that your clients would actually ignore the `access_url` entirely? If so, I'm missing something -- how do they know from where to fetch the object bytes?

rdeborj:
@dglazer yes to the former. Objects would be fetched using auth credentials obtained via their own institute identity provider.  I should have been more specific when I wrote we would use ```access_url``` for information purposes only, my apologies.

dglazer:
Thanks for the quick response @rdeborj -- that usage pattern makes perfect sense to me, and I expect other implementations with org-specific auth needs will do something similar.

dglazer:
@rishidev @briandoconnor -- given that it's now more than 10 days since Rishi reached out to driver projects (see https://github.com/ga4gh/data-repository-service-schemas/pull/236#issuecomment-472416164), and there are no blocking concerns, are we okay to merge this PR?

briandoconnor:
Ah, I accidentally moved this around in the project board.  I'm putting it in review now and Rishi will merge soon once he annotates with the drivers that have responded/not responded.

rishidev:
0 vote input over email from Genomics England (Mark Brundrett), 
+1 from ENA/EVA/EGA (Jorge),
+0 from TopMed (Albert Smith)

PR #237, opened 2019-03-04
Release/0.0.1

briandoconnor:
This is a merge of the develop branch as it was right at the transition from DOS to DRS.  So it includes the updated documentation and renames of our DRS work but the full set of DOS endpoints that we may cherrypick from in the future.  I'm merging with master here and then will tag as a 0.0.1 ahead of our 0.1 DRS release so we can always easily refer back to our starting point.


dglazer:
Pinging the reviewers @david4096 @natanlao @jaeddy -- any comments or concerns? (I *think* this is purely historical housekeeping -- important as a future reference, but doesn't affect the ongoing DRS 0.1 discussion.)

briandoconnor:
OK, I've heard no objection for several weeks and this is just housekeeping. Since 0.1 is right around the corner I'm merging this to get this moved forward and to be able to tag a 0.0.1 release from master before a bunch of 0.1 PRs get merged

Issue #242, opened 2019-04-01
Refine return object schema to remove protobuf artifacts


delagoya:
## Background

We should think about removing the GetXXXRequest/and GetXXXResponse object schema return type envelopes, as these are a hold-over from when the schema was defined in Protocol Buffers and PB needed these constructs. OpenAPI response schema objects do not need these, as evidenced by the TRS schema:

```yaml
basePath: /ga4gh/trs/v2
tags:
  - name: GA4GH
    description: A set of resources proposed as a common standard for tool repositories
paths:
  '/tools/{id}':
    get:
      summary: 'List one specific tool, acts as an anchor for self references'
      operationId: toolsIdGet
      description: >-
        This endpoint returns one specific tool (which has ToolVersions nested
        inside it)
     parameters:
        - name: id
          in: path
          required: true
          type: string
          description: >-
            A unique identifier of the tool, scoped to this registry, for
            example `123456`
      responses:
        '200':
          description: A tool.
          schema:
            $ref: '#/definitions/Tool'
        '404':
          description: The tool can not be found.
          schema:
            $ref: '#/definitions/Error'
```

Here is an example from the current DRS schema:

```yaml
paths:
  '/objects/{object_id}':
    get:
      summary: Retrieve a Data Object
      operationId: GetObject
      responses:
        '200':
          description: The Data Object was found successfully.
          schema:
            $ref: '#/definitions/GetObjectResponse'
# ...
definitions:
  GetObjectResponse:
    type: object
    required: ['object']
    properties:
      object:
        $ref: '#/definitions/Object'
```

And an example of the proposed change:

```yaml
paths:
  '/objects/{object_id}':
    get:
      summary: Retrieve a Data Object
      operationId: GetObject
      responses:
        '200':
          description: The Data Object was found successfully.
          schema:
            $ref: '#/definitions/Object'
# ...
definitions:
    type: object
    required: ['id', 'size', 'created', 'checksums', 'access_methods']
    properties:
      id:
        type: string
        description: |-
          An identifier unique to this Data Object.
      name:
        type: string
        description: |-
          A string that can be optionally used to name a Data Object.
# ...
```

This naming scheme **is also inconsistently used** in the schema, and sometimes it is used to define an object directly. Example:

```yaml
paths:
  '/service-info':
    get:
      summary: Returns service version and other information
      operationId: GetServiceInfo
      responses:
        '200':
          description: Service information returned successfully
          schema:
[...]

PR #243, opened 2019-04-01
Removed artifacts from protobuf translation within response object schema

delagoya:
* ErrorResponse -> Error
  * ServiceInfoResponse -> ServiceInfo
  * /bundles/{bundle_id}#get reference GetBundleResponse -> Bundle
  * /objects/{object_id}#get reference GetObjectResponse -> Object
  * Removed GetBundleResponse and GetObjectResponse envelope objects
  * GetAccessURLResponse -> AccessURL

Addresses Issue #242

### DIFF
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -30,7 +30,7 @@ paths:
         '200':
           description: Service information returned successfully
           schema:
-            $ref: '#/definitions/ServiceInfoResponse'
+            $ref: '#/definitions/ServiceInfo'
       tags:
         - DataRepositoryService
       x-swagger-router-controller: ga4gh.drs.server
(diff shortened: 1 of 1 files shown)


dglazer:
@delagoya , this is looking good to me. (And doing a quick check of the WES schema, it's already halfway between the old and new models -- it still has `ErrorResponse` and `RunListResponse`, but all the other methods just return what they need to return.)

Do you have more changes in mind, or do you think this covers it?
And do you have any concerns about the way it landed, or are you liking it?

delagoya:
It would be good if the Error object could be standardized across the API's, but open to suggestions on how to implement this. May be worth it's own issue & PR

Looks like @susheel addressed the `etag` vs  `multipart-md5` issue and some dangling unused schema objects in PR #246 

Any other schema changes are unrelated to this issue, so I would propose those go in another issue + PR.

dglazer:
Got it. Looking a bit more closely:
- Get{Bundle,Object,AccessURL}Response are the meat of this PR, and I agree they simplify things, so LGTM.
- ErrorResponse and ServiceInfoResponse are just renames; they don't change the API structure at all, and the new names are a little cleaner, so LGTM also. I agree we should standardize across our APIs, but don't feel the need to do that for 0.1

@delagoya -- sounds like we should push this forward -- please add @briandoconnor and me as official reviewers if you agree, so we can LGTM officially and add it to the list we want the driver projects to review.

delagoya:
@dglazer  I do not have permission to assign reviewers (or anything else other than submit the issue and PR). @rishidev?

briandoconnor:
LGTM, and I agree to push future changes to the PR from Susheel and also future PRs.  I'm giving my thumbs up on this PR in the mean time

ohofmann:
+1 from AGHA.

junjun-zhang:
+1 from ICGC-ARGO

susheel:
+1 ELIXIR

Issue #221, opened 2019-01-29
Define Collection methods


rishidev:
Some points to discuss here:
How similar/different to object methods?
Distinguished endpoints for each or different?

briandoconnor:
@susheel

tetron:
From the perspective of CWL / WES, one thing that needs be clarified is how a DRS collection is intended to be materialized to a directory.  When a DRS collection is used as input to a workflow:

* How are filenames assigned?  Does the collection assign its own pathnames, or does it use the "name" field from the data object, or some combination of both?
* What happens if the "name" field of the data object is not defined (since it's currently labeled as optional)?
* What happens if there is a name collision?
* Can a single collection describe a file hierarchy (such as assigning pathnames that may include '/' to each file), or are subdirectories only represented as sub-collections?
** If sub-collections, how are they named?  What happens if there is a name collision between a data object and a collection?
* Can sub-collections be shared between collections?
* Does read access to a collection imply read access to data objects and/or sub-collections contained within?
* Can empty subdirectories be represented?

delagoya:
Thanks @tetron, this is a good list to start with. Before we discuss each point, I want to confirm some agreements that were proposed during the Boston meeting:

1. That collections are immutable
2. Collections are not versioned
3. That you can have sub-collections
4. That a `GET` operation on a collection returns the immediate children (e.g. sub-collections require additional API operations to unroll).

tetron:
1. yes
2. yes

3 & 4. that's what we agreed to, but I don't think we fully explored the alternatives.

tetron:
This is the representation of a bundle in the current spec:

```
object_ids: [
  "drs://object1",
  "drs://object2",
  "drs://bundle3",
]
```

In order to materialize these into a directory, you would use the `name` field of each object or bundle.  In order for the collection to be immutable, the `name` field of each data object or bundle would also have to be immutable.  Which means if you want to rename a data object, you have to create a copy.

An alternate representation, where the collection assigns names to objects:

```
contents: {
  "/foo.bam": "drs://object1",
  "/foo.bai": "drs://object2",
  "/subdir/bar.bam": "drs://object3",
  "/subdir/bar.bai": "drs://object4"
}
```

Instead of sub-collections, a collection represents the entire directory tree.  This eliminates the possibility of circular dependencies by collections including themselves.  Fetching the collection record gives you more information up front about the contents of the collection, without having to fetch each object record separately.

susheel:
[...]

PR #244, opened 2019-04-02
Data Bundles v0.1 Schema (1/2)

susheel:
Initial commit based on #221.

Advanced expansions i.e. `expand=shallow` & `expand=deep` will have to wait until OAPI3 (for anyOf support)

[Optional] Data Bundle Access URL Method added in #245

### DIFF
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -258,7 +258,6 @@ definitions:
       - id
       - size
       - created
-      - updated
       - version
       - checksums
       - contents
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -42,7 +42,7 @@ paths:
         '200':
           description: Successfully found the Data Bundle.
           schema:
-            $ref: '#/definitions/GetBundleResponse'
+            $ref: '#/definitions/Bundle'
         '400':
           description: The request is malformed.
           schema:
@@ -51,12 +51,12 @@ paths:
           description: The request is unauthorized.
           schema:
             $ref: '#/definitions/ErrorResponse'
-        '404':
-          description: The requested Data Bundle wasn't found.
-          schema:
-            $ref: '#/definitions/ErrorResponse'
         '403':
           description: The requester is not authorized to perform this action.
+          schema:
+            $ref: '#/definitions/ErrorResponse'
+        '404':
+          description: The requested Data Bundle wasn't found.
           schema:
             $ref: '#/definitions/ErrorResponse'
         '500':
@@ -194,57 +194,70 @@ definitions:
           sha512
   Bundle:
     type: object
-    required: ['id', 'object_ids', 'created', 'updated', 'version', 'checksums']
     properties:
       id:
         type: string
-        description: |-
+        description: >-
           An identifier, unique to this Data Bundle
       object_ids:
         type: array
         items:
           type: string
-        description: |-
+        description: >-
           The list of Data Objects that this Data Bundle contains.
       created:
         type: string
         format: date-time
-        description: |-
+        description: >-
           Timestamp of object creation in RFC3339.
       updated:
         type: string
         format: date-time
-        description: |-
-          Timestamp of update in RFC3339, identical to create timestamp in systems
-          that do not support updates.
+        description: >-
+          Timestamp of update in RFC3339, identical to create timestamp in
+          systems that do not support updates.
       version:
         type: string
-        description: |-
-          A string representing a version, some systems may use checksum, a RFC3339
-          timestamp, or incrementing version number. For systems that do not support
-          versioning please use your update timestamp as your version.
+        description: >-
+          A string representing a version, some systems may use checksum, a
+          RFC3339 timestamp, or incrementing version number. For systems that do not
+          support versioning please use your update timestamp as your version.
+      mime-type:
+        type: string
+        description: >-
+          A string providing the mime-type of the Data Bundle.
+          For example, "application/json".
       checksums:
         type: array
-        items:
-          $ref: '#/definitions/Checksum'
-        description: |-
+        description: >-
           At least one checksum must be provided.
           The Data Bundle checksum is computed over all the checksums of the
           Data Objects that bundle contains.
-      description:
-        type: string
-        description: |-
-          A human readable description.
-      aliases:
+        items:
+          $ref: '#/definitions/Checksum'
+      access_methods:
         type: array
         items:
+          $ref: '#/definitions/AccessMethod'
+      description:
+        type: string
+        description: A human readable description.
+      aliases:
+        type: array
+        description: A list of strings that can be used to identify this Data Bundle.
+        items:
           type: string
-        description: |-
-          A list of strings that can be used to identify this Data Bundle.
-      system_metadata:
-        $ref: '#/definitions/SystemMetadata'
-      user_metadata:
-        $ref: '#/definitions/UserMetadata'
+      contents:
+        type: array
+        items:
+          $ref: '#/definitions/BundleObjects'
+    required:
+      - id
+      - size
+      - created
+      - updated
+      - version
+      - checksums
   Object:
     type: object
     required: ['id', 'size', 'created', 'checksums', 'access_methods']
@@ -308,11 +321,6 @@ definitions:
           These aliases can be used to represent the Data Object's location in
           a directory (e.g. "bucket/folder/file.name") to make Data Objects
           more discoverable. They might also be used to represent
-  GetBundleResponse:
-    type: object
-    properties:
-      bundle:
-        $ref: '#/definitions/Bundle'
   GetObjectResponse:
     type: object
     required: ['object']
@@ -396,3 +404,15 @@ definitions:
       license:
         type: object
         description: License information for the exposed API
+  BundleObjects:
+    type: object
+    properties:
+      name:
+        type: string
+      id:
+        type: string
+    required:
+      - name
+      - id
+tags:
+  - name: DataRepositoryService
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -216,7 +216,7 @@ definitions:
         type: string
         description: >-
           A string providing the mime-type of the Data Bundle.
-          For example, "application/json".
+        example: 'application/json'
       checksums:
         type: array
         description: >-
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -222,7 +222,7 @@ definitions:
         description: >-
           At least one checksum must be provided.
           The Data Bundle checksum is computed over a sorted concatenation of all the 
-          checksums of the Data Objects that bundle contains. The list of Data
+          checksums within the top-level 'contents' of the Bundle (not recursive). The list of Data
           Object checksums are sorted alphabetically (hex-code) before concatenation and a further
           checksum performed on the concated binary value. Example below:
           Data Ojects:
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -418,8 +418,8 @@ definitions:
       id:
         type: string
         description: >-
-          This represents the DRS identifier that may point to a Data Object
-          or another Data Bundle that will need to be recursed.
+          This represents the DRS identifier URI that may point to a Data
+          Object or another Data Bundle that will need to be recursed.
       type:
         type: string
         enum:
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -186,6 +186,11 @@ definitions:
         type: string
         description: >-
           An identifier, unique to this Data Bundle
+      size:
+        type: string
+        format: int64
+        description: |-
+          The computed cumulated size of all Data Objects in bytes.
       created:
         type: string
         format: date-time
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -160,19 +160,6 @@ securityDefinitions:
     name: Authorization
     in: header
 definitions:
-  SystemMetadata:
-    type: object
-    additionalProperties: true
-    description: |-
-            OPTIONAL
-            These values are reported by the underlying object store.
-            A set of key-value pairs that represent system metadata about the object.
-  UserMetadata:
-    type: object
-    additionalProperties: true
-    description: |-
-            OPTIONAL
-            A set of key-value pairs that represent metadata provided by the uploader.
   Checksum:
     type: object
     required: ['checksum']
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -194,7 +194,7 @@ definitions:
         type: string
         format: int64
         description: |-
-          The cumulative size of all (recursive) Data Objects in bytes.
+          The cumulative size of all Data Objects and Bundles (within contents) in bytes.
       created:
         type: string
         format: date-time
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -418,8 +418,15 @@ definitions:
       id:
         type: string
         description: >-
-          This represents the DRS identifier URI that may point to a Data
+          This represents the DRS identifier that may point to a Data
           Object or another Data Bundle that will need to be recursed.
+      drs_uri:
+        type: array
+        description: >-
+          This represents a list of suggested full DRS identifier URI paths
+          that may used obtain the Data Object or Data Bundle that will need
+          to be recursed. These URIs may be external to this DRS.
+        example: 'drs://example.com/ga4gh/drs/v1/objects/{object_id}'
       type:
         type: string
         enum:
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -419,8 +419,18 @@ definitions:
         description: >-
           This represents the DRS identifier that may point to a Data Object
           or another Data Bundle that will need to be recursed.
+      type:
+        type: string
+        enum:
+          - object
+          - bundle
+        description: >-
+          This represents the type of DRS object referenced within this
+          Data Bundle. BundleObject of type bundle will need to be recursed
+          further.
     required:
       - name
       - id
+      - type
 tags:
   - name: DataRepositoryService
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -194,7 +194,8 @@ definitions:
         type: string
         format: int64
         description: |-
-          The cumulative size of all Data Objects and Bundles (within contents) in bytes.
+          The cumulative size of all Data Objects and Bundles listed within
+          contents field in bytes.
       created:
         type: string
         format: date-time
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -253,7 +253,7 @@ definitions:
         description: >-
           The list of Data Objects that this Data Bundle contains.
         items:
-          $ref: '#/definitions/BundleObjects'
+          $ref: '#/definitions/BundleObject'
     required:
       - id
       - size
@@ -406,7 +406,7 @@ definitions:
       license:
         type: object
         description: License information for the exposed API
-  BundleObjects:
+  BundleObject:
     type: object
     properties:
       name:
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -238,7 +238,11 @@ definitions:
         description: A human readable description.
       aliases:
         type: array
-        description: A list of strings that can be used to identify this Data Bundle.
+        description: |-
+          A list of strings that can be used to find other metadata 
+          about this Data Bundle from external metadata sources. These
+          aliases can be used to represent the Data Bundle's secondary
+          accession numbers or external GUIDs.
         items:
           type: string
       contents:
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -409,6 +409,11 @@ definitions:
     properties:
       name:
         type: string
+        description: >-
+          This represents the Bundle author declared name that must be
+          used when materialising the associated data object and will 
+          override any name directly associated with the object itself.
+          This string MUST NOT contain any slashes.
       id:
         type: string
     required:
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -199,12 +199,6 @@ definitions:
         type: string
         description: >-
           An identifier, unique to this Data Bundle
-      object_ids:
-        type: array
-        items:
-          type: string
-        description: >-
-          The list of Data Objects that this Data Bundle contains.
       created:
         type: string
         format: date-time
@@ -249,6 +243,8 @@ definitions:
           type: string
       contents:
         type: array
+        description: >-
+          The list of Data Objects that this Data Bundle contains.
         items:
           $ref: '#/definitions/BundleObjects'
     required:
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -258,7 +258,6 @@ definitions:
       - id
       - size
       - created
-      - version
       - checksums
       - contents
   Object:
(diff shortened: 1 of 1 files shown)


openapi/data_repository_service.swagger.yaml
```diff
@@ -194,57 +194,71 @@ definitions:
...
         type: string
-        description: |-
-          A string representing a version, some systems may use checksum, a RFC3339
-          timestamp, or incrementing version number. For systems that do not support
-          versioning please use your update timestamp as your version.
+        description: >-
+          A string representing a version, some systems may use checksum, a
+          RFC3339 timestamp, or incrementing version number. For systems that do not
+          support versioning please use your update timestamp as your version.
+      mime-type:
```
dglazer: I don't think  `mime-type` makes sense on bundles -- aren't they always just a collection of other objects? (The referenced objects can of course be typed, but that's different.)

openapi/data_repository_service.swagger.yaml
```diff
@@ -194,57 +194,71 @@ definitions:
...
+          For example, "application/json".
       checksums:
         type: array
-        items:
-          $ref: '#/definitions/Checksum'
-        description: |-
+        description: >-
           At least one checksum must be provided.
           The Data Bundle checksum is computed over all the checksums of the
           Data Objects that bundle contains.
```
dglazer: Can we be more explicit about how to compute the checksum?
* Is the idea simple checksum string concatenation? E.g. if Bundle B1 contains Objects O1 and O2, with checksums "cs111" and "cs222", then do I compute the checksum of bundle B1 by running my hash function on "cs111cs222"?
* what order should I use in computing my checksum? E.g. if Bundle B2 contains the same two objects, but happens to return them in a different order in the `contents` array, does it have a different checksum? 
* does the checksum include the `name` field from `BundleObjects`? If not, are we okay with two bundles that return differently named objects having the same checksum, as long as the objects have the same contents?

openapi/data_repository_service.swagger.yaml
```diff
@@ -194,57 +194,71 @@ definitions:
...
+        items:
+          $ref: '#/definitions/Checksum'
+      access_methods:
```
dglazer: Do we need `access_methods` for Bundles? Given that the payload of a bundle is the `contents` array, which is returned in the response to this method, what further access is needed?

openapi/data_repository_service.swagger.yaml
```diff
@@ -396,3 +405,15 @@ definitions:
       license:
         type: object
         description: License information for the exposed API
+  BundleObjects:
+    type: object
+    properties:
+      name:
+        type: string
```
dglazer: please add a description (since it's not obvious that this string represents the name that should be used when materializing the object, and that the name here overrides any name that might be associated with the object itself)

openapi/data_repository_service.swagger.yaml
```diff
@@ -396,3 +405,15 @@ definitions:
...
+      id:
+        type: string
```
dglazer: please add a description, clarifying:
1. that this is the ID of a DRS object
2. whether it can also be the ID of a DRS bundle (and if so, how I know which is which)

openapi/data_repository_service.swagger.yaml
```diff
@@ -194,57 +194,71 @@ definitions:
...
-        $ref: '#/definitions/SystemMetadata'
-      user_metadata:
-        $ref: '#/definitions/UserMetadata'
+      contents:
+        type: array
+        items:
+          $ref: '#/definitions/BundleObjects'
+    required:
+      - id
+      - size
```
dglazer: this object doesn't have a `size` field

openapi/data_repository_service.swagger.yaml
```diff
@@ -194,57 +194,71 @@ definitions:
...
+      - created
+      - updated
```
dglazer: this shouldn't be required (to match Object)

openapi/data_repository_service.swagger.yaml
```diff
@@ -194,57 +194,71 @@ definitions:
...
+      - version
```
dglazer: this shouldn't be required (to match Object)

openapi/data_repository_service.swagger.yaml
```diff
@@ -194,57 +194,71 @@ definitions:
...
   Bundle:
     type: object
-    required: ['id', 'object_ids', 'created', 'updated', 'version', 'checksums']
     properties:
       id:
         type: string
-        description: |-
+        description: >-
           An identifier, unique to this Data Bundle
       object_ids:
```
dglazer: isn't `object_ids` redundant, now that we have `contents` ?

openapi/data_repository_service.swagger.yaml
dglazer: also -- should we be explicit that this string is not allowed to contain slashes?

openapi/data_repository_service.swagger.yaml
susheel: Good catch. I forgot to remove it. Removed in 03c1b01ba49227f8f15737b8a282876e40946f58

openapi/data_repository_service.swagger.yaml
susheel: Description updated. The algorithm is based on a data bundle checksum is based on a concatenated alphabetically (hex-code) sorted checksum of data objects

openapi/data_repository_service.swagger.yaml
susheel: Yes, I think so. If the entire List of Data Objects in a Bundle existed in a FTP/S3/... directory, we would want to provide a `access url` to the whole directory. It would also help requesting access to the whole directory via a single `access_id`.

If such a directory does not exist, then the DRS provider simply provides an empty list of `access_methods`.

openapi/data_repository_service.swagger.yaml
susheel: If you are able to access the data in a bundle via an `access_url` then mime-type may be important. It also allows future extensions for container formats that can be accessed as a bundle and or individually.

openapi/data_repository_service.swagger.yaml
susheel: I missed this from the object. It should have a `size` computed from the cumulative sizes of each Data Object

openapi/data_repository_service.swagger.yaml
susheel: We need to have either `created/updated` or `version` to let the user know if they are accessing the latest version or an older version of the bundle.

openapi/data_repository_service.swagger.yaml
susheel: We need to have either `created/updated` or `version` to let the user know if they are accessing the latest version or an older version of the bundle.

openapi/data_repository_service.swagger.yaml
susheel: Description added.

openapi/data_repository_service.swagger.yaml
susheel: Description added. I have added `type` to the BundleObject to help the user make this distinction.

openapi/data_repository_service.swagger.yaml
dglazer: Hmm -- that means all implementors would need to compute the total size for all objects in a bundle before they could expose it via DRS. I guess that's okay, given that we also require them to compute the checksum, but I'd like to hear from at least a couple people planning to implement Bundle support before imposing that mandate.

openapi/data_repository_service.swagger.yaml
dglazer: That feels like it's taking us down the slippery slope of being opinionated about versioning, which we were trying to avoid for now. I suggest leaving `updated` as optional, and then starting a separate conversation in a new issue/PR about whether we want to make it required for **both** `Object` and `Bundle`. (Unless there's some argument for them being different? I don't see one).

openapi/data_repository_service.swagger.yaml
dglazer: Same comment as on `updated` -- I suggest making it optional for now, and then separately discussing a change to both `Object` and `Bundle`.

openapi/data_repository_service.swagger.yaml
```diff
@@ -194,57 +181,87 @@ definitions:
...
-          versioning please use your update timestamp as your version.
+        description: >-
+          A string representing a version, some systems may use checksum, a
+          RFC3339 timestamp, or incrementing version number. For systems that do not
+          support versioning please use your update timestamp as your version.
+      mime-type:
+        type: string
+        description: >-
+          A string providing the mime-type of the Data Bundle.
+          For example, "application/json".
```
dglazer: I suggest using the `example:` YAML syntax for consistency, and slightly friendlier documentation formatting

openapi/data_repository_service.swagger.yaml
susheel: Okay. Moving `updated` to optional

openapi/data_repository_service.swagger.yaml
susheel: Okay, I agree

openapi/data_repository_service.swagger.yaml
```diff
@@ -194,57 +181,85 @@ definitions:
...
+      name:
+        type: string
         description: |-
-          The list of Data Objects that this Data Bundle contains.
+          A string that can be optionally used to name a Data Bundle.
+      size:
+        type: string
+        format: int64
+        description: |-
+          The computed cumulated size of all Data Objects in bytes.
```
dglazer: Is this recursive? In other words, if the Bundle contains Bundles that contain Objects, is `size` expected to be based on walking the whole tree? Either way, let's clarify the description. (And nit -- I think "computed cumulated" should be "cumulative" or "total".)

openapi/data_repository_service.swagger.yaml
```diff
@@ -194,57 +181,85 @@ definitions:
...
+        type: string
+        description: >-
+          A string providing the mime-type of the Data Bundle.
+          For example, "application/json".
       checksums:
         type: array
+        description: >-
+          At least one checksum must be provided.
+          The Data Bundle checksum is computed over a sorted concatenation of all the 
+          checksums of the Data Objects that bundle contains. The list of Data
```
dglazer: Similar comment as on `size` -- is this recursive, meaning it's based on the full list of Objects all the way down? Or is it just based on the checksums of the top-level Objects and Bundles?

dglazer: Confirming that the checksum does *not* include the `name` field? If so, that means two Bundles that contain the same Objects, but laid out in completely different directory trees, would have the same checksum -- is that intentional? (Not sure what makes sense here.)

openapi/data_repository_service.swagger.yaml
```diff
@@ -396,3 +406,33 @@ definitions:
       license:
         type: object
         description: License information for the exposed API
+  BundleObjects:
```
dglazer: I suggest `BundleObject` (singular), since it's only one thing.

dglazer:
For those following along -- there are several small refinements being discussed and resolved in line-by-line comments. A few of those comments (in particular the comment threads on `mime-type` and `access_methods`, and the related new method in #245) feel like they're part of a higher-level discussion of how a bundle is expected to be fetched.

* My original vision was a Bundle was simply a possibly-nested list of Objects, meant to be fetched iteratively using our existing mechanisms. 
* I could also easily imagine someone storing a logical collection of files inside a single Object, possibly compressed, and meant to be fetched monolithically using our existing mechanisms and then unpacked by the client.
* We *could* also support a hybrid, where a single Bundle can be fetched either iteratively or monolithically. I don't know if any of our drivers need that, but if so, I'd suggest a simpler implementation, where the Bundle has a new optional field (e.g `bundle_object_id`) that allows callers who don't want to iterate to use the existing methods to fetch it one go. That way our Bundle methods stay focused on the iterative case, and don't need to duplicate all the monolithic Object fetch cases.

openapi/data_repository_service.swagger.yaml
```diff
@@ -194,57 +181,85 @@ definitions:
...
+          RFC3339 timestamp, or incrementing version number. For systems that do not
+          support versioning please use your update timestamp as your version.
+      mime-type:
+        type: string
+        description: >-
+          A string providing the mime-type of the Data Bundle.
+          For example, "application/json".
       checksums:
         type: array
+        description: >-
```
ddietterich: Why is this an array here, but a single value for Object? That seems inconsistent.
I think it should be a simple reference to the `Checksum` with the computation method as described below.

openapi/data_repository_service.swagger.yaml
ddietterich: I don't understand @susheel. You have stored the bundle in a file. But instead of producing that file contents as the response to this request, you want the client to make another request that will produce the Bundle response? So then you are not producing the contents array below?

openapi/data_repository_service.swagger.yaml
```diff
@@ -194,57 +181,85 @@ definitions:
...
+        items:
+          $ref: '#/definitions/AccessMethod'
       description:
         type: string
-        description: |-
-          A human readable description.
+        description: A human readable description.
       aliases:
         type: array
+        description: A list of strings that can be used to identify this Data Bundle.
```
ddietterich: Probably need to incorporate the changes to align with our discussion in the meeting today.

openapi/data_repository_service.swagger.yaml
```diff
@@ -396,3 +406,33 @@ definitions:
...
+        type: string
+        description: >-
+          This represents the Bundle author declared name that must be
+          used when materialising the associated data object and will 
+          override any name directly associated with the object itself.
+          This string MUST NOT contain any slashes.
+      id:
+        type: string
+        description: >-
+          This represents the DRS identifier that may point to a Data Object
```
ddietterich: Is DRS identifier well-defined somewhere? (Easily could be. I'm still finding my way around here...) I assume this is intended to be a complete DRS URI - not just the id component. Is that correct?

openapi/data_repository_service.swagger.yaml
susheel: @dglazer Recursive

@ddietterich Providing this would give the user a handle to the total size of the Bundle they will be downloading. If having `access_methods` are included in the Bundle specification then having `size` is a must.

openapi/data_repository_service.swagger.yaml
susheel: @dglazer Just the top-level, i.e. contents

susheel: @dglazer Yes, two Bundles with the same list of objects (in any order) should have the same checksum.

This would allow the user to lay out the same physical (byte) Bundles in different (named) ways based on their specific usecase.

openapi/data_repository_service.swagger.yaml
susheel: Updated.

openapi/data_repository_service.swagger.yaml
susheel: @ddietterich No, @dglazer's question is more about should Bundles have a`ccess_methods`. I believe the specification should not restrict this.
The `access_url` of a Bundle provides a way to access the Bundle `contents` byte array in a single call. Consider a folder example:
ftp://server.com/data/A
A
├── DO1
└── DO2
The Data Objects `DO1` & `DO2` will all have access_urls:
DO1.*.access_url: ftp://server.com/data/A/DO1
DO2.*.access_url: ftp://server.com/data/A/DO2

A client would hand these `access_url`s to ftp-client for download.

If I were to create a Bundle A with it's `contents` represented by `DO1` and `DO2`, then the question is should I be able to provide an access_url for the Bundle. My argument is that the specification should not restrict this use case.
A.access_url: ftp://server.com/data/A

A client would hand this `access_url` to ftp-client for download.

The service provider must be certain that directory structure provided by the `access_url` conforms to the structure described by the Bundle `contents`

openapi/data_repository_service.swagger.yaml
susheel: Updated

openapi/data_repository_service.swagger.yaml
```diff
@@ -194,57 +181,85 @@ definitions:
...
           An identifier, unique to this Data Bundle
-      object_ids:
-        type: array
-        items:
-          type: string
+      name:
+        type: string
         description: |-
-          The list of Data Objects that this Data Bundle contains.
+          A string that can be optionally used to name a Data Bundle.
```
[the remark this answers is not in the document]
susheel: @ddietterich It should be implied by the missing `required` tag. No harm in reiterating this in the attribute description. This is mainly to make sure users don't use the name as the Bundle or Object's primary identifier.

I think the naming scheme will be implementation-specific.

openapi/data_repository_service.swagger.yaml
```diff
@@ -194,57 +181,85 @@ definitions:
...
+      size:
+        type: string
+        format: int64
+        description: |-
+          The computed cumulated size of all Data Objects in bytes.
       created:
         type: string
         format: date-time
-        description: |-
+        description: >-
```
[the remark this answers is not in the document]
susheel: @ddietterich Yes Bundle creation and update time, NOT of its contents.

openapi/data_repository_service.swagger.yaml
```diff
@@ -194,57 +181,85 @@ definitions:
...
+          Timestamp of update in RFC3339, identical to create timestamp in
+          systems that do not support updates.
       version:
         type: string
-        description: |-
-          A string representing a version, some systems may use checksum, a RFC3339
-          timestamp, or incrementing version number. For systems that do not support
-          versioning please use your update timestamp as your version.
+        description: >-
```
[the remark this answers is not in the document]
susheel: > If I make a new version of the directory, does it get a new create time? In which case, the create and update time are the same?

Yes.

> Why would you ask non-versioning systems to use the update timestamp for a version? That seems arbitrary. I'd prefer to use something that matches what I plan to use for versions if/when we do them.

It was a remnant from the older Bundle description. I have updated the description.

openapi/data_repository_service.swagger.yaml
susheel: Yes, it should be the complete DRS URI. I will clarify the description.

I think the DRS ID URI must be defined somewhere in the documentation.

openapi/data_repository_service.swagger.yaml
susheel: @ddietterich A Bundle or Object can have an array of checksums calculated using any hashing algorithm, not just MD5. The Checksum object provides the value and type of algorithm used.

dglazer:
@ddietterich -- that's my prior also.
Susheel, are you picturing a use case equivalent to `zip -r filename.zip folder` as an optional alternate way to fetch the directory tree?

openapi/data_repository_service.swagger.yaml
```diff
@@ -194,57 +181,84 @@ definitions:
...
+          A string providing the mime-type of the Data Bundle.
+        example: 'application/json'
       checksums:
         type: array
+        description: >-
+          At least one checksum must be provided.
+          The Data Bundle checksum is computed over a sorted concatenation of all the 
+          checksums within the top-level 'contents' of the Bundle (not recursive). The list of Data
+          Object checksums are sorted alphabetically (hex-code) before concatenation and a further
```
dglazer: As written, the checksum only includes the checksums of the top-level Objects -- is that right? I would have expected it to also include the checksums of any top-level Bundles. Maybe just replace "Data Object checksums" with "Object or Bundle checksums"?

openapi/data_repository_service.swagger.yaml
dglazer: See https://github.com/ga4gh/data-repository-service-schemas/pull/244#issuecomment-480517411

openapi/data_repository_service.swagger.yaml
```diff
@@ -396,3 +405,33 @@ definitions:
...
+        type: string
+        description: >-
+          This represents the Bundle author declared name that must be
+          used when materialising the associated data object and will 
+          override any name directly associated with the object itself.
+          This string MUST NOT contain any slashes.
+      id:
+        type: string
+        description: >-
+          This represents the DRS identifier URI that may point to a Data
```
dglazer: Is this a URI? I expected this to be an `id` string that can be passed to `GET /objects/{object_id}` or `GET /bundles/{bundle_id}`.

briandoconnor:
@susheel: Idea from conversation today, move the access_methods from this PR to a new PR to simplify this one.

He's also going to clarify documentation for URIs for objects rather than ID on its own.

openapi/data_repository_service.swagger.yaml
susheel: Removed Bundle `mime-type` from this PR. Will place it in a different PR for discussion.

openapi/data_repository_service.swagger.yaml
susheel: Removed `access_methods` to be discussed in a different PR.

openapi/data_repository_service.swagger.yaml
dglazer: And to align with our further discussion in the meeting **today**.

openapi/data_repository_service.swagger.yaml
susheel: Clarified Bundle > size in description. size represents only the cumulative size of all objects within the Bundle's contents field (so depth 1)

openapi/data_repository_service.swagger.yaml
susheel: Tx. Description updated.

openapi/data_repository_service.swagger.yaml
susheel: BundleObject updated with a `drs_uri` that represents a list of possible URIs to obtain the data object or bundle.

susheel:
@dglazer Cleaned up the PR based on today's call:
- Added `drs_uri` field to BundleObject (represents a list of possible DRS URI paths to obtain the object or bundle
- Removed `mime-type` and `access_methods` for a separate discussion
- Clarified Bundle > `size` and Bundle > `checksum` descriptions.

openapi/data_repository_service.swagger.yaml
dglazer: @ddietterich I think it's an array for `Object` also? (I agree with you they should match; I think they already do.)

openapi/data_repository_service.swagger.yaml
dglazer: Or, since this PR doesn't actually change the text of the description, you could just remove the syntax change to the YAML, and then this edit won't conflict with the one in #246 -- you call which is easier.

openapi/data_repository_service.swagger.yaml
```diff
@@ -194,57 +181,77 @@ definitions:
...
-          versioning please use your update timestamp as your version.
+        description: >-
+          A string representing a version, some systems may use checksum, a
+          RFC3339 timestamp, or incrementing version number.
       checksums:
         type: array
+        description: >-
+          At least one checksum must be provided.
+          The Data Bundle checksum is computed over a sorted concatenation of all 
+          the checksums within the top-level 'contents' of the Bundle (not recursive).
```
dglazer: @susheel -- confirming that this checksum does **not** include the `name` field from the `contents` array? (I"m not sure which makes more sense -- as written it means that two bundles that contain the same objects, but with different names, will end up having the same checksum as each other.)

dglazer:
@susheel , this is looking very close. Two unresolved comments in the code:
- confirm that Bundle `checksums` do **not** take `name` fields into account
- fix a merge conflict in the description of `aliases` (between this PR and #246)

openapi/data_repository_service.swagger.yaml
susheel: Clarified description to confirm names are NOT included in the checksum computation. Yes, two bundles containing the same objects should resolve to have the same checksum. The names assigned by the user should not affect the checksum, just how the individual objects are materialised onto disk.

openapi/data_repository_service.swagger.yaml
susheel: I changed the `aliases` description to align with #246

openapi/data_repository_service.swagger.yaml
```diff
@@ -396,3 +403,40 @@ definitions:
...
+          used when materialising the associated data object and will 
+          override any name directly associated with the object itself.
+          This string MUST NOT contain any slashes.
+      id:
+        type: string
+        description: >-
+          This represents the DRS identifier that may point to a Data
+          Object or another Data Bundle that will need to be recursed.
+      drs_uri:
+        type: array
```
ddietterich: I think the array needs an
```items:
    type: string
```

susheel: Thanks @ddietterich. Resolved.

PR #246, opened 2019-04-02
Metadata cleanup

susheel:
- Merge #243 for proto2openapi cleanup
- Renamed `multipart-md5` to `etag`
- Removed unused `SystemMetadata` and `UserMetadata` definitions

### DIFF
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -226,7 +226,10 @@ definitions:
         items:
           type: string
         description: |-
-          A list of strings that can be used to identify this Data Bundle.
+          A list of strings that can be used to find this Data Bundle.
+          These aliases can be used to represent the Data Bundle's 
+          secondary accession numbers or external GUIDs that may help
+          with the discovery of the Data Bundle.
   Object:
     type: object
     required: ['id', 'size', 'created', 'checksums', 'access_methods']
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -159,19 +159,6 @@ securityDefinitions:
     name: Authorization
     in: header
 definitions:
-  SystemMetadata:
-    type: object
-    additionalProperties: true
-    description: |-
-            OPTIONAL
-            These values are reported by the underlying object store.
-            A set of key-value pairs that represent system metadata about the object.
-  UserMetadata:
-    type: object
-    additionalProperties: true
-    description: |-
-            OPTIONAL
-            A set of key-value pairs that represent metadata provided by the uploader.
   Checksum:
     type: object
     required: ['checksum']
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -161,12 +161,12 @@ securityDefinitions:
 definitions:
   Checksum:
     type: object
-    required: ['checksum']
+    required:
+      - checksum
     properties:
       checksum:
         type: string
-        description: |-
-          The hex-string encoded checksum for the Data.
+        description: 'The hex-string encoded checksum for the data'
       type:
         type: string
         description: |-
@@ -364,7 +364,8 @@ definitions:
         description: The integer representing the HTTP status code (e.g. 200, 404).
   ServiceInfoResponse:
     type: object
-    required: ['version']
+    required:
+      - version
     description: Placeholder for the Info Object
     properties:
       version:
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -292,7 +292,7 @@ definitions:
           A list of strings that can be used to find this Data Object.
           These aliases can be used to represent the Data Object's location in
           a directory (e.g. "bucket/folder/file.name") to make Data Objects
-          more discoverable. They might also be used to represent
+          more discoverable.
   GetBundleResponse:
     type: object
     properties:
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -289,10 +289,10 @@ definitions:
         items:
           type: string
         description: |-
-          A list of strings that can be used to find this Data Object.
-          These aliases can be used to represent the Data Object's location in
-          a directory (e.g. "bucket/folder/file.name") to make Data Objects
-          more discoverable.
+          A list of strings that can be used to find other metadata 
+          about this Data Object from external metadata sources. These
+          aliases can be used to represent the Data Object's secondary
+          accession numbers or external GUIDs.
   GetBundleResponse:
     type: object
     properties:
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -227,10 +227,6 @@ definitions:
           type: string
         description: |-
           A list of strings that can be used to identify this Data Bundle.
-      system_metadata:
-        $ref: '#/definitions/SystemMetadata'
-      user_metadata:
-        $ref: '#/definitions/UserMetadata'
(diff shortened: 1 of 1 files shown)


openapi/data_repository_service.swagger.yaml
```diff
@@ -160,36 +159,23 @@ securityDefinitions:
     name: Authorization
     in: header
 definitions:
-  SystemMetadata:
```
dglazer: SystemMetadata and UserMetadata are still being referenced in the Bundle definition, so the documentation build is failing

susheel: References are removed in PR #244, lines [244](https://github.com/ga4gh/data-repository-service-schemas/pull/244/files#diff-4bd72006051046bcde21e06812a2c330L244) and [246](https://github.com/ga4gh/data-repository-service-schemas/pull/244/files#diff-4bd72006051046bcde21e06812a2c330L246)

dglazer: Got it -- unfortunately splitting it across PRs means that this one fails the Travis checks. Do you plan to merge them somehow, or would you like me to review some other PRs first, so those can be merged and then we can come back to this one?

susheel: I'll remove the references here so that the build passes and this could be merged without conflict. Once merged PR #244 should go through as it makes the same deletions.

dglazer:
@susheel , thanks for pulling this PR together. if I'm reading correctly, it does three things:
1) replace `multipart-md5` with `etag` as an optional checksum type <-- this is the main point
2) remove `SystemMetadata` and `UserMetadata` from the bundle type <-- this is logically part of #244; I'd slightly prefer to move it there, so the `Bundle` discussion can happen in one place, but I'm fine doing it here if you prefer. (I fully support the removal, whichever PR it lives in.)
3) misc small formatting changes to the YAML (e.g. sorting error codes)

Questons:
- is that right, or did I miss other substantive changes?
- does this now include all the metadata changes you think are appropriate for v0.1?
- re-reading the `Object` fields, my one open question is about `aliases` -- it looks like their description field is incomplete -- do you understand the intended end-to-end flow enough to flesh that out? And if not, should we delete it from 0.1, and add it back when we have an end-to-end use case?

susheel:
@dglazer 
> 1. replace `multipart-md5` with `etag` as an optional checksum type <-- this is the main point

Yes

> 2. remove `SystemMetadata` and `UserMetadata` from the bundle type <-- this is logically part of #244; I'd slightly prefer to move it there, so the `Bundle` discussion can happen in one place, but I'm fine doing it here if you prefer. (I fully support the removal, whichever PR it lives in.)

I've removed it in both. Hopefully, they won't conflict when they both get merged.


> Questons:
> * is that right, or did I miss other substantive changes?

No, nothing else.

> * does this now include all the metadata changes you think are appropriate for v0.1?

yes.

> * re-reading the `Object` fields, my one open question is about `aliases` -- it looks like their description field is incomplete -- do you understand the intended end-to-end flow enough to flesh that out? And if not, should we delete it from 0.1, and add it back when we have an end-to-end use case?

Yes, We'd like to use it as secondary accession numbers/urls for looking up further information about the Bundle/Object. Ideally, we would like to build a @context scheme for this, but I think we can leave this is in to allow users to explore its utility.

dglazer:
Thanks Susheel. All sounds good, with an open question on `aliases`. The current description is:
> A list of strings that can be used to find this Data Object. These aliases can be used to represent the Data Object's location in a directory (e.g. "bucket/folder/file.name") to make Data Objects more discoverable. They might also be used to represent

That's incomplete and unclear, and sounds different from your vision. Can you propose a new description in this PR, so we can make sure driver projects know what they're agreeing to? A few thoughts in advance of seeing that:
- I get your idea of accession numbers -- maybe we should just simplify it to that, and give an example with a doi: string (or similar)?
- if it's meant to help with object search and discovery, I'd leave it out until we have a plan with that workgroup
- if it's meant to map to directories, it sounds like it conflicts with the Bundles model

openapi/data_repository_service.swagger.yaml
```diff
@@ -51,14 +50,14 @@ paths:
...
             $ref: '#/definitions/ErrorResponse'
-        '404':
-          description: The requested Data Bundle wasn't found.
-          schema:
-            $ref: '#/definitions/ErrorResponse'
         '403':
           description: The requester is not authorized to perform this action.
           schema:
             $ref: '#/definitions/ErrorResponse'
+        '404':
```
briandoconnor: great!  How did we miss that!?

openapi/data_repository_service.swagger.yaml
```diff
@@ -160,36 +159,23 @@ securityDefinitions:
...
         type: string
         description: |-
           The digest method used to create the checksum. If left unspecified md5
           will be assumed.
 
           possible values:
-          md5                # most blob stores provide a checksum using this
-          multipart-md5      # multipart uploads provide a specialized tag in S3
+          md5               # most blob stores provide a checksum using this
+          etag              # multipart uploads to blob stores
```
briandoconnor: is this a common term across clouds?  Is there a better place to define this in the docs vs. a comment?

briandoconnor:
Alias:
* SBG: 1) used aliases for internal ids and GUIDs, 2) another DRS URL
* STAGE/TOPMed: used for GUIDs
* JADE (All of Us): not used right now

How are these searched/looked up by?

Can we move the alias conversation to another PR?  

Let's clean up the text here for this PR for now and then address aliases in another PR.

briandoconnor:
@susheel can you please help clean up the text on alias?

openapi/data_repository_service.swagger.yaml
susheel: Not really. AFAIK the ETags specification does not specify the algorithm to be used. Service provider and users are free to choose what they use to put in the ETag header.

Typically the major cloud providers support multipart-md5. The trouble is that it depends a lot on how you have chunked your file (default is 16Mb)

dglazer:
@susheel , I'm ready to approve as soon as we clean up the `aliases` description.

susheel:
@dglazer @briandoconnor Update Bundle alias description.

junjun-zhang:
ICGC-ARGO +1

At ICGC, alias is data repository specific. For example, we transferred a BAM file with ID `EGAF00000866152` from `EGA` repository to `Collab` repository we assigned a new ID `98a9b481-2a22-571a-97ce-46e3b8ddecb1` (`EGAF00000866152` becomes an alias).

openapi/data_repository_service.swagger.yaml
```diff
@@ -240,11 +226,10 @@ definitions:
         items:
           type: string
         description: |-
-          A list of strings that can be used to identify this Data Bundle.
-      system_metadata:
-        $ref: '#/definitions/SystemMetadata'
-      user_metadata:
-        $ref: '#/definitions/UserMetadata'
+          A list of strings that can be used to find this Data Bundle.
```
dglazer: I'm concerned about saying "can be used to find", since this API has no search features, and we don't yet have integration with the search API.

openapi/data_repository_service.swagger.yaml
```diff
@@ -307,7 +292,7 @@ definitions:
           A list of strings that can be used to find this Data Object.
```
dglazer: @susheel -- I like your edits to the Bundle `aliases` field definition. Can you make the same edits to the Object `aliases` field?

susheel: Done.

dglazer: I don't see any edits? I'll see if I can go ahead and make the change myself to save back and forth.

susheel:
LGTM

Issue #239, opened 2019-03-13
Providing more metadata than access_url with GetAccessURLResponse?


briandoconnor:
This is a followup to our conversation on 20190311 and PR #236

See https://github.com/ga4gh/data-repository-service-schemas/pull/236#discussion_r264309098

`GetAccessURLResponse` just has `access_url` in the response.  Do we want to expand that to include more metadata in the future so DRS implementors can have the option to pass back more info than just a URL?  This would be needed if DRS implementors want to use DRS to hand back enough information to "get the bytes" for a variety of protocols beyond URLs/signed URLs.

sarpera:
_Sort of related to this, adding `authorization_metadata` property to the items in `access_methods` array was also discussed in [this issue](https://github.com/ga4gh/data-repository-service-schemas/issues/47#issuecomment-467446033) as well._

**So far:**
`GetAccessURLResponse` via `/objects/<object_id>/access/<access_id>` is meant to return a ready-to-use byte access URL (e.g signed URL) already. Client is assumed to already have chosen an access method from the `access_methods` array and passed its `access_id` to above path to (optionally with auth token) get a url to bytes. 

**Suggestion:**
We can extend this for the cases like aspera, globus etc where `access_url` needs to be used along with some additional information like temp. credentials. I think it still doesn't defeat the original purpose; separating object metadata and access url that requires additional steps (signing, providing temp. credentials etc) to provide. 

Quoting from the the [discussion](https://github.com/ga4gh/data-repository-service-schemas/pull/236#discussion_r264309098) PR 236:

> In the previous response that came back we returned a URL along with system_metadata, user_metadata, and authorization_metadata. In the NIH Data Commons we needed the ability to provide more context about the URL to allow a user to make a request. 

Some examples how we can extend the original PR based on the use cases: 

Assumptions:
- Both `/objects/<id>` and  `/objects/<object_id>/access/<access_id>` are optionally behind some auth mechanism. We can easily add BasicAuth, 0Auth2 etc in `securityDefinitions` so that all standard auth mechanisms are allowed in DRS and calls to DRS are not ambiguous.
- Providing `access_url` may be costly and therefore unnecessary to provide each time an object method is requested
- Object metadata might be publicly accessible, but access url might require some auth

**Use Case 1**

- DRS contains an object that is stored in a private-access storage system
- Client doesn't have direct access to storage system; action required to get access (sign, get credentials/access token etc)
- A URL alone is enough to provide byte-access

[...]

PR #248, opened 2019-04-03
allow optional headers on access_url (resolves issue #239)

dglazer:
This PR resolves issues #239. Changes:
1) introduce a new `AccessURL` object that contains a `url` string and an optional `headers` array
2) replace the `access_url` string with the new object, in `AccessMethod` and in the `/access` method
3) small documentation cleanup

Staged documentation: https://ga4gh.github.io/data-repository-service-schemas/preview/feature/issue-239-access-credentials/docs/

### DIFF
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -122,7 +122,7 @@ paths:
         '200':
           description: The access URL was found successfully.
           schema:
-            $ref: '#/definitions/GetAccessURLResponse'
+            $ref: '#/definitions/AccessURL'
         '400':
           description: The request is malformed.
           schema:
@@ -325,12 +325,6 @@ definitions:
     properties:
       object:
         $ref: '#/definitions/Object'
-  GetAccessURLResponse:
-    type: object
-    required: ['object']
-    properties:
-      access_url:
-        $ref: '#/definitions/AccessURL'
   AccessURL:
     type: object
     required: ['url']
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -330,8 +330,21 @@ definitions:
     required: ['object']
     properties:
       access_url:
+        $ref: '#/definitions/AccessURL'
+  AccessURL:
+    type: object
+    required: ['url']
+    properties:
+      url:
         type: string
         description: A fully resolvable access_url that can be used to fetch the actual object bytes.
+      headers:
+        type: object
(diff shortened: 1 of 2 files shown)


dglazer:
@kemp-google @philloooo -- fyi -- this PR allows DRS servers to optionally return auth headers along with access URLs. It's meant to finish the plan discussed in #239 -- comments welcome

briandoconnor:
FYI, the swagger can be found here: https://ga4gh.github.io/data-repository-service-schemas/preview/feature/issue-239-access-credentials/swagger-ui/#/DataRepositoryService

openapi/data_repository_service.swagger.yaml
```diff
@@ -319,13 +325,22 @@ definitions:
...
+      url:
         type: string
-        description: A fully resolvable access_url that can be used to fetch the actual object bytes.
+        description: A fully resolvable URL that can be used to fetch the actual object bytes.
+      headers:
+        type: array
+        items:
+          type: string
+        description: >-
+          An optional list of headers to include in the HTTP request to `url`.
```
briandoconnor: I think this is a great first step.  It improves what we have right now and we should go for it.  As an implementer/user of DRS I'd like to know 1) what are the possible headers here, do we define them in the /service-info end point for example or in the spec as a list of acceptable ones?  And 2) it would be great to see examples of how these work for each of the types of access methods DRS 1.0 would support (in the docs for example).  Totally fine to address in a future PR

dglazer: Re (1): As a user, you shouldn't have to know what headers are possible; you just use the ones you're given. As an implementor, you can return any headers you want. The most likely use case is to return an Authorization header, as is shown in the example in the YAML.

Re (2): this PR already includes beefed-up documentation on access methods (in front_matter.adoc); happy to add more in future PRs

briandoconnor:
See what the Drivers think after @sarpera takes a look and gives thumbs up.  Does it work for others using things like aspera or globus

briandoconnor:
Notes from call: @jorizci mentioned expiration of tokens... will work well for static system but not sure how this will work systems that expire their tokens quickly.

dglazer:
Summarizing conversation with @jorizci on the workstream call -- this PR is not sufficient to fully meet EGA's use case, but he's a "-0", not a "-1". From his perspective, it's okay to merge, *and* we should discuss further enhancements in Hinxton.

junjun-zhang:
+1 from ICGC-ARGO

Issue #241, opened 2019-03-25
Confirm DRS implementations for April meeting


briandoconnor:
Followup with implementers for DRS to see April plans:
* Elixir / Susheel - will be implementing by April
* Illumina - not for April meeting
* Google / Aaron - not by April but could make a proof of concept on GCP if there's a shortage of implementations for April.
* Verily/Broad / DanD - hope to have for April
* Seven Bridges - not by April
* HCA - maybe (Brian)
* TOPMed - checking on this (Abby)

Also, it would be really helpful to find out what access methods each will support.  See https://github.com/ga4gh/data-repository-service-schemas/issues/240 and comment on what you will be supporting/wanting to support in 2019

susheel:
@briandoconnor ELIXIR will be implementing a DRS server by April.

mattions:
@briandoconnor Seven Bridges will implement a DRS client and a DRS server on the CGC, Cavatica and DataStage.

The three different platforms will implement this in different timings, with the the Cancer Genomics Cloud going first.

@rishidev the Cancer Genomics Cloud is one of the NCI Cancer Cloud Resources, so it represents a part of that driver project.

The same applies for the DataStage platform, which is part of the TOPMed driver project.

Cavatica is integrated in the Gabriella Miller's Kids First, a project form the NCI Common fund. It's not an official driver project, but it is active in the group in the same way, so I guess can represent external stakeholder.

ddietterich:
@briandoconnor We are in the process of implementing our DRS server. We hope to have something to show in April; it is on the cusp.

briandoconnor:
Illumina not for April meeting

briandoconnor:
By April SBG won't have it done, by June is their target.

kemp-google:
At the moment, we do not plan to have an implementation by the meetings in April.  However, if there is a shortage of implementations, we could likely produce a simple one that backs onto GCS in that time frame.

briandoconnor:
Would be good to get the implementers to comment on #240 to find out what access methods are needed by each implementation

Issue #240, opened 2019-03-13
refine the list of AccessMethod type values


dglazer:
Following up on the initial `type` list for `AccessMethod`, as enumerated in #236:
- how do we want to support  htsget? My first thought is it should be a `type` = `htsget` of its own, where the `access_url` is required to support the htsget protocol, vs. just being a place to fetch bytes.
- do we want our initial `type` list to only include methods that we expect somebody to have implemented by v1 launch, or should we lean further into the hypothetical future? I'm inclined to trim it back to only things that will be live in 2019, once we know what that list is.
- can we combine `http` and `https` into one? I expect most servers will only use one of them, and as a client I don't think I would ever have interesting logic to pick one vs. the other -- if I intend to use HTTP, I'll pick that `type` and use whatever `access_url` I'm given.

briandoconnor:
Suggestion: Let's take out the access methods that aren't in use by implementers.

briandoconnor:
So here's the current list:

- s3
- gs
- ftp
- sftp
- http
- https
- nfs
- globus
- aspera
- gsiftp
- local

I *think* we need the following for 2019:

- s3
- gs
- https
- htsget
- local

But what do implementers need?

briandoconnor:
See #249 , the PR for the changes proposed here.  Let's keep that PR feature branch in sync with what the drivers/implementers here need in 2019

susheel:
@briandoconnor For ELIXIR we need:
- local [MUST]
- ftp [MUST]
- globus [MUST]
- gsiftp [MUST]
- http(s) [NICE]
- s3 [NICE]
- htsget [NICE]

ohofmann:
Vote for s3 and htsget from the AGHA side (not an implementer though, mind you).

mattions:
@briandoconnor from SBG we are looking at these for our use-cases:

- s3 [MUST]
- gs [MUST]
- https [NICE]
- htsget [MUST]
- local [NICE]

ddietterich:
For Jade Data Repo:
- gs [MUST]
- s3 [MUST]

briandoconnor:
Awesome!  Thanks so much folks for the quick reply.  I updated the feature branch.

delagoya:
I would suggest using `file` instead of `local` since this is an official [URI scheme](https://tools.ietf.org/html/rfc8089)

briandoconnor:
OK, @delagoya I made that change

jorizci:
[...]

PR #249, opened 2019-04-08
Reduced down the list of access methods

briandoconnor:
See #240, it would be extremely helpful to use that issue to collect what access methods are needed by each driver/implementation.  And we can propagate any changes to the list of access methods via this feature branch before merging.

See also issue #241 for the list of known DRS implementations.

### DIFF
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -336,15 +336,9 @@ definitions:
         enum:
         - s3
         - gs
-        - ftp
-        - sftp
-        - http
         - https
-        - nfs
-        - globus
-        - aspera
-        - gsiftp
         - local
+        - htsget
         description: >-
           Type of the access method.
       access_url:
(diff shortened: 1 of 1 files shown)


briandoconnor:
See https://ga4gh.github.io/data-repository-service-schemas/preview/feature/issue-240-list-access-methods/swagger-ui/#/

briandoconnor:
OK, I got feedback from quite a few drivers/implementers.  @dglazer can you take a look?

dglazer:
@shaikh-rashid -- can you say a bit more about CanDIG's requirements? How do you plan to retrieve data objects if you're not using an `access_method` ?

junjun-zhang:
+1 from ICGC-ARGO

jorizci:
+1 From EGA

susheel:
+1 ELIXIR

briandoconnor:
@sarpera brings up htsget vs https, needed for htsget?

@delagoya brings up that there are additional methods supported by htsget that clients need to know about.

@jorizci will take a look to make sure this understanding is correct

susheel:
htsget - https://samtools.github.io/hts-specs/htsget.html

PR #254, opened 2019-04-24
Corrected Bundle Size as integer

### LINKED ISSUE #253: Bundle object size is defined as a string
This is a cut'n'paste of the problem I found in #251.
The Bundle says:
```
      size:
        type: string
        format: int64
```
but it should be `type: integer`

susheel:
fixes #253


susheel:
Deleting in the hope that #258 fixes this.

Issue #252, opened 2019-04-22
bundle or object in URI convention


dvoet:
the [work-in-progress docs](https://ga4gh.github.io/data-repository-service-schemas/preview/develop/docs/#_uri_convention_work_in_progress) for URI convention suggest the form `drs://<server>/<id>`. But this does not let the caller know if the identified thing is an object or a bundle making it hard to figure out the right request to make. I suggest adding something in the path to distinguish. `drs://<server>/<bundle|object>/<id>` or `drs://<server>/<b|o>/<id>` if length is a concern.

briandoconnor:
Good point, I reached the same conclusion when cleaning up the docs for v0.1.  @dvoet can you take a look at release 0.1 (https://ga4gh.github.io/data-repository-service-schemas/preview/release/0.1/docs/) and see what you think?  If folks don't like the style I put together here then we can open a PR and get something better in for v0.2.  

Another thing, the spec does not cover is how to use GUIDs of various flavors with DRS (e.g. DOIs, https://dataguids.org/, etc).  We don't attempt to create a standard around this or conventions for this in DRS at this time.  So a URI like drs://<GUID> is out of scope (but, again, we could either add it in scope after talking to our drivers and add it in via a PR to 0.2 or say it's out of scope and something we should partner on with Discovery).

briandoconnor:
I talked with @dglazer and we came to the conclusion that, since this is not full baked, we should pull the text about URIs from the current spec docs for v0.1.

So here's that text that outlines one way of doing it:

For convenience, we define a recommended syntax for fully referencing DRS-accessible objects. Strings of the form drs://<server>/<path>/<[bundle|object]>/<endpoint>/<id> mean “make a DRS
GET to the HTTPS address at <server>, passing in the DRS id <id> at the end of the appropriate <path>/<[bundle|object]>/<endpoint>, to retrieve the DRS JSON response”.
The <endpoint> is optional depending on what you are doing e.g. the <endpoint> is /access when retrieving a particular access method.  This is all just
a complicated way of just saying "replace drs:// with https:// and GET on the URL that results."

For example:

* drs://test.drs.org/ga4gh/drs/v1/bundles/5007fc21-6430-4fb9-82b3-520c60e26751 -> maps to a GET request to https://test.drs.org/ga4gh/drs/v1/bundles/5007fc21-6430-4fb9-82b3-520c60e26751
[...]

PR #258, opened 2019-04-29
 Object / Bundle unification (addresses part of #252)

dglazer:
Changes, as discussed in https://github.com/ga4gh/data-repository-service-schemas/issues/252#issuecomment-487535043: 
- make Object.access_methods optional (but document that it's required for Objects)
- add an optional Object.contents field (but document that it's required for Bundles)
- delete the Bundle object and the /bundles method

Still needs work on documentation, to consistently refer to objects, blobs, and bundles.

### DIFF
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -241,9 +241,9 @@ definitions:
         description: >-
           If not set, this `Object` is a single blob.
           
-          If set, this `Object` is a bundle containing the listed `BundleObject` s (some of which may be further nested).
+          If set, this `Object` is a bundle containing the listed `ContentsObject` s (some of which may be further nested).
         items:
-          $ref: '#/definitions/BundleObject'
+          $ref: '#/definitions/ContentsObject'
       description:
         type: string
         description: |-
@@ -341,7 +341,7 @@ definitions:
       license:
         type: object
         description: License information for the exposed API
-  BundleObject:
+  ContentsObject:
     type: object
     properties:
       name:
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -198,23 +198,34 @@ definitions:
         minItems: 1
         items:
           $ref: '#/definitions/Checksum'
-        description: |-
+        description: >-
           The checksum of the `Object`. At least one checksum must be provided.
+          
           For blobs, the checksum is computed over the bytes in the blob.
+          
           
           For bundles, the checksum is computed over a sorted concatenation of the 
           checksums of its top-level contained objects (not recursive, names not included).
           The list of checksums is sorted alphabetically (hex-code) before concatenation
           and a further checksum is performed on the concatenated checksum value.
           
+          
           For example, if a bundle contains blobs with the following checksums:
+          
             md5(blob1) = 72794b6d30bc86d92e40a1aa65c880b8
-            md5(blob2) = 5e089d29a18954e68a78ee6a3c6edabd       
+            
+            md5(blob2) = 5e089d29a18954e68a78ee6a3c6edabd   
+            
           Then the checksum of the bundle is:
+          
             md5( concat( sort( md5(blob1), md5(blob2) ) ) )
+            
               = md5( concat( sort( 72794b6d30bc86d92e40a1aa65c880b8, 5e089d29a18954e68a78ee6a3c6edabd ) ) )
+              
               = md5( concat( 5e089d29a18954e68a78ee6a3c6edabd, 72794b6d30bc86d92e40a1aa65c880b8 ) )
+              
               = md5( 5e089d29a18954e68a78ee6a3c6edabd72794b6d30bc86d92e40a1aa65c880b8 )
+              
               = f7a29a0422e7d870b10839ad6c985079
       access_methods:
         type: array
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -35,45 +35,6 @@ paths:
       tags:
         - DataRepositoryService
       x-swagger-router-controller: ga4gh.drs.server
-  '/bundles/{bundle_id}':
-    get:
-      summary: Get info about a Data Bundle.
-      description: >-
-        Returns bundle metadata, and a list of ids that can be used to fetch bundle contents.
-      operationId: GetBundle
-      responses:
-        '200':
-          description: The Data Bundle was found successfully.
-          schema:
-            $ref: '#/definitions/Bundle'
-        '400':
-          description: The request is malformed.
-          schema:
-            $ref: '#/definitions/Error'
-        '401':
-          description: The request is unauthorized.
-          schema:
-            $ref: '#/definitions/Error'
-        '403':
-          description: The requester is not authorized to perform this action.
-          schema:
-            $ref: '#/definitions/Error'
-        '404':
-          description: The requested Data Bundle wasn't found.
-          schema:
-            $ref: '#/definitions/Error'
-        '500':
-          description: An unexpected error occurred.
-          schema:
-            $ref: '#/definitions/Error'
-      parameters:
-        - name: bundle_id
-          in: path
-          required: true
-          type: string
-      tags:
-        - DataRepositoryService
-      x-swagger-router-controller: ga4gh.drs.server
   '/objects/{object_id}':
     get:
       summary: Get info about a Data Object.
@@ -101,7 +62,6 @@ paths:
           description: The requested Data Object wasn't found
           schema:
             $ref: '#/definitions/Error'
-
         '500':
           description: An unexpected error occurred.
           schema:
@@ -192,83 +152,9 @@ definitions:
           etag              # multipart uploads to blob stores
           sha256
           sha512
-  Bundle:
-    type: object
-    required: ['id', 'size', 'created', 'checksums', 'contents']
-    properties:
-      id:
-        type: string
-        description: >-
-          An identifier unique to this Data Bundle.
-      name:
-        type: string
-        description: >-
-          A string that can be used to name a Data Bundle.
-      size:
-        type: string
-        format: int64
-        description: >-
-          The cumulative size, in bytes, of all Data Objects and Bundles listed in
-          the `contents` field.
-      created:
-        type: string
-        format: date-time
-        description: >-
-          Timestamp of Bundle creation in RFC3339.
-      updated:
-        type: string
-        format: date-time
-        description: >-
-          Timestamp of Bundle update in RFC3339, identical to create timestamp in
-          systems that do not support updates.
-      version:
-        type: string
-        description: >-
-          A string representing a version. (Some systems may use checksum, a
-          RFC3339 timestamp, or an incrementing version number.)
-      checksums:
-        type: array
-        description: |-
-          The checksum of the Data Bundle. At least one checksum must be provided.
-          
-          The Data Bundle checksum is computed over a sorted concatenation of all 
-          the checksums (names not included) within the top-level 'contents' of the
-          Bundle (not recursive). The list of Data Object or Bundle checksums are 
-          sorted alphabetically (hex-code) before concatenation and a further checksum
-          is performed on the concatenated checksum value.
-          Example below:
-          Data Ojects:
-            md5(DO1) = 72794b6d30bc86d92e40a1aa65c880b8
-            md5(DO2) = 5e089d29a18954e68a78ee6a3c6edabd
-          Data Bundle:
-          DB1 = md5( concat( sort( md5(DO1), md5(DO2) ) ) )
-              = md5( concat( sort( 72794b6d30bc86d92e40a1aa65c880b8, 5e089d29a18954e68a78ee6a3c6edabd ) ) )
-              = md5( concat( 5e089d29a18954e68a78ee6a3c6edabd, 72794b6d30bc86d92e40a1aa65c880b8 ) )
-              = md5( 5e089d29a18954e68a78ee6a3c6edabd72794b6d30bc86d92e40a1aa65c880b8 )
-              = f7a29a0422e7d870b10839ad6c985079
-        items:
-          $ref: '#/definitions/Checksum'
-      description:
-        type: string
-        description: A human readable description of the Data Bundle.
-      aliases:
-        type: array
-        description: >-
-          A list of strings that can be used to find other metadata 
-          about this Data Bundle from external metadata sources. These
-          aliases can be used to represent the Data Bundle's secondary
-          accession numbers or external GUIDs.
-        items:
-          type: string
-      contents:
-        type: array
-        description: >-
-          The list of Data Objects and Data Bundles contained by this Data Bundle.
-        items:
-          $ref: '#/definitions/BundleObject'
   Object:
     type: object
-    required: ['id', 'size', 'created', 'checksums', 'access_methods']
+    required: ['id', 'size', 'created', 'checksums']
     properties:
       id:
         type: string
@@ -282,7 +168,8 @@ definitions:
         type: integer
         format: int64
         description: |-
-          The object size in bytes.
+          For single items, the object size in bytes.
+          For bundles, the cumulative size, in bytes, of everything listed in the `contents` field.
       created:
         type: string
         format: date-time
@@ -292,7 +179,7 @@ definitions:
         type: string
         format: date-time
         description: >-
-          Timestamp of Object update in RFC3339, identical to create timestamp in systems
+          Timestamp of object update in RFC3339, identical to create timestamp in systems
           that do not support updates.
       version:
         type: string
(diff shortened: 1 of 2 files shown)


openapi/data_repository_service.swagger.yaml
```diff
@@ -301,34 +188,63 @@ definitions:
       mime_type:
         type: string
         description: |-
-          A string providing the mime-type of the Data Object.
+          A string providing the mime-type of the `Object`.
         example:
           application/json
       checksums:
         type: array
```
susheel: Add `minItems : 1` to ensure at least one checksum is provided.

dglazer: Done -- pls resolve if it LGTY

dglazer:
For easy reference: [HTML documentation](https://ga4gh.github.io/data-repository-service-schemas/preview/feature/issue-252-object-bundle-merge/docs/), [Swagger UI](https://ga4gh.github.io/data-repository-service-schemas/preview/feature/issue-252-object-bundle-merge/swagger-ui/#/DataRepositoryService/)

sarpera:
Based on the swagger changes, this may lead to ambiguous responses from a DRS server i.e the schema cannot enforce a blob (file) not to have "contents" part. This example response where an ID that returns metadata about a blob (file) with "contents" i.e list of other blobs / bundles becomes a valid response:

GET drs://example.com/ga4gh/drs/v1/objects/123
```yaml
{
  "id": "123",
  "name": "file.bam",
  "size": 4567,
  "created": "2019-05-09T11:01:26.625Z",
  "updated": "2019-05-09T11:01:26.625Z",
  "checksums": [
    {
      "checksum": "123",
      "type": "md5"
    }
  ],
  "access_methods": [
    {
      "type": "s3",
      "access_url": {
        "url": "s3://foo/file.bam",
        "headers": {
          "Authorization": "Basic Z2E0Z2g6ZHJz"
        }
      },
      "region": "us-east-1"
    }
  ],
  "contents": [
    {
      "name": "file2.bam",
      "id": "124",
      "drs_uri": "drs://example.com/ga4gh/drs/v1/objects/124"
    }
  ]
}
``` 

Additionally to above remark, IMO responses from a REST API should not be interpreted differently (blob vs bundle) based on attributes in the response while parsing it.

openapi/data_repository_service.swagger.yaml
```diff
@@ -192,97 +152,24 @@ definitions:
...
-      contents:
-        type: array
-        description: >-
-          The list of Data Objects and Data Bundles contained by this Data Bundle.
-        items:
-          $ref: '#/definitions/BundleObject'
   Object:
     type: object
-    required: ['id', 'size', 'created', 'checksums', 'access_methods']
+    required: ['id', 'size', 'created', 'checksums']
```
susheel: @dglazer To help address @sarpera's comment, I think both `contents` and `access_methods` must be required even if they are empty.

openapi/data_repository_service.swagger.yaml
```diff
@@ -192,97 +152,24 @@ definitions:
...
-          type: string
-      contents:
-        type: array
-        description: >-
-          The list of Data Objects and Data Bundles contained by this Data Bundle.
-        items:
-          $ref: '#/definitions/BundleObject'
   Object:
     type: object
-    required: ['id', 'size', 'created', 'checksums', 'access_methods']
```
susheel: @dglazer I forget the reason why `access_methods` was removed from `object`

junjun-zhang:
Reading through the discussions on object/bundle unification, I feel it's getting to really like a file system. Objects are files and bundles are folders, it's also mentioned there can be multiple levels of nested bundles.

This is not a bad thing at all, if DRS specs allows (or at least does not preclude) implementers resemble Linux file system, it would make it much easier to integrate with common command line tools. Something like this `samtools view drs://my-drs-server/awesome-project/myreads.bam chr1:10-10000` could work out-of-the-box.

There are many file/data object management systems already working this way, here are a few examples:
* Google bucket browser, try this: https://console.cloud.google.com/storage/browser/genomics-public-data/1000-genomes/. You can navigate through folder hierarchy up/down and download files directly. As we all know, there is actually no folder exists. Slashes are just part of the object ID to mimic folders. When you download a file you are redirected to a GET request with URL like this: `/download/storage/v1/b/genomics-public-data/o/1000-genomes%2FREADME`, notice the string after `/o/` is the actual object ID and it's URL encoded with `/` converted to `%2F`.
* Synapse, eg: https://www.synapse.org/#!Synapse:syn9732876
* Seven Bridges' *Files* module of the CGC system
* Dropbox, see screenshot below:
image

Although they're all implemented in different ways, all of them provide similar user experience: data objects are organized similar to a file system.

This approach does not only unify Object and Bundle discussed here, but also supports secondary files discussed in #260. Because a DRS implementer can mimic file system behaviour, it would be much the same as how secondary file is supported in a native file system, eg, `drs://my-drs-server/awesome-project/myreads.bam.bai` as the index file of the BAM file mentioned earlier.

This sounds cool, but what about GUID and all the nice features come with it? like decentralized ID generation, immutability for permanent referencing etc. With careful design of DRS, it's possible to get the best out of both worlds: user-friendly file system like behaviour and GUID.

Here is one possibility. We use object `name` to support file system like behaviour and use `id` for non-user-friendly global unique identifier. The DRS specs is probably already written in such way. It may look like this:
```
[
  {
    "id": "b8e12273-dcf1-42ff-9818-118b9b976b01",
    "name": "",   # 'root' bundle, perhaps with empty name?
    "contents": [
      {
        "id": "fea8f84d-516e-478c-b65d-6cab3db2aa67"
      }
    ],
    "parent_bundle": null  # this is 'root' bundle, no parent bundle
  },
  {
    "id": "fea8f84d-516e-478c-b65d-6cab3db2aa67",
    "name": "awesome-project",
    "contents": [
      {
        "id": "3bc755df-ce87-437e-ac58-b972af79cb0a"
      },
      {
        "id": "cf329985-453b-4088-92e7-773a37e16020"
      }
    ],
    "parent_bundle": "b8e12273-dcf1-42ff-9818-118b9b976b01"  #  links back to 'root' bundle
  },
  {
    "id": "3bc755df-ce87-437e-ac58-b972af79cb0a",
    "name": "myreads.bam",
    "parent_bundle": "fea8f84d-516e-478c-b65d-6cab3db2aa67"  # suggesting this for two-way linking to enable better client-side exploration
  },
  {
    "id": "cf329985-453b-4088-92e7-773a37e16020",
    "name": "myreads.bam.bai",
    "aliases": [ "myreads.bai" ],  # another name of the bai file
    "parent_bundle": "fea8f84d-516e-478c-b65d-6cab3db2aa67"
  },
]
```

With this, it's possible to address any DRS object in two ways (note the addition of `name` and `id` in the URI):
* by name: `drs://my-drs-server/name/awesome-project/myreads.bam`. This is fully qualified name with bundles separated by slashes.
* by id: `drs://my-drs-server/id/3bc755df-ce87-437e-ac58-b972af79cb0a`

To make it work unambiguously, `/` may not be in a `name`. However having `/` in `id` would be just fine. We should also document that `name` is not stable may change while `id` is immutable once assigned.

I am not suggesting specific fields and their names here, just want to use the example here to illustrate the ideas.

There are potentially other benefits as well:
* provide clients the capability to explore a DRS server starting from `drs://my-drs-server/name/`, which is the `root`. This immediately makes DRS useful given that `v1` specs does not consider search capabilities
* perhaps in a DRS prototype, one can fairly easily to map local file system, gs or s3 buckets with file path like object IDs to DRS objects/bundles

Lastly, just want to point out another system we all use shares some similarity with what described above, that is the *git* version control system. Everything *git* stores is an *object*, it has four types: `blob`, `tree`, `commit` and `tag`. The first two are equivalent to DRS's `blob` and `bundle`. Git gives user experience much like using a file system, but internally it keeps everything in a flat list of data objects (with content hash code being the IDs, naturally immutable). Maybe one day when DRS needs to support object versioning, some ideas from *git* could be borrowed.

susheel:
@junjun-zhang In ELIXIR we plan to use hashes as our decentralised identifiers - mainly inspired by git, ipfs and dat. As you have correctly identified we plan to use objects and bundles as a mirror for git blob and tree. Our current plan is to hash one more time to obtain a version hash similar to git commits. We are just missing one more field for objects/bundles for representing the previous version to complete our view of we'd like DRS to be implemented in ELIXIR.

We also lean towards using slashes in object/bundle IDs, precisely for the reasons to note - as a slash is just a char in the object is. There is a point for not using it as it would break the currently proposed `/access` API, this can be fixed - I hope

junjun-zhang:
@susheel, glad that we are on the same page.
Regarding `access` endpoint, what you proposed in https://github.com/ga4gh/data-repository-service-schemas/issues/263#issue-439640607 should work. It's actually better in my view: `GET /access/{access_id}` is simpler than `GET /objects/{object_id}/access/{access_id}`.

briandoconnor:
Does the unification of the GET for bundles and objects break RESTful-ness of our service?  My understanding is REST offers entities that you can operate on with verbs such as GET, POST, and PUT.  But this case our entities are really polymorphic.  Is that OK with folks?  Or does that start to break down the clarity about what you're working with from a DRS server?

How will this work when we eventually add POST/PUT endpoints?  Will they be unified or will we have different endpoints since creating a bundle entity is different from creating an object?

openapi/data_repository_service.swagger.yaml
dglazer: Because they're not required for bundles. (This is one of the small downsides of a single method. IMO, that's more than made up for by the upside of having DRS IDs be self-contained.)

openapi/data_repository_service.swagger.yaml
dglazer: Hmm -- I'm not sure why requiring an optionally-empty array is better than having an optional array. Both require the client to be aware of what's going on, so what's the benefit of requiring the server to generate often-empty placeholder elements? (Open to answers, but I don't see one.)

dglazer:
@sarpera , re ambiguous responses containing both `contents` and `access-methods` -- I believe @susheel wants that to be legit, so he can return a bundle that you can either fetch in pieces (by iterating through the `contents`) or all at once (e.g. by downloading a .zip file).

@sarpera @briandoconnor , re tradeoffs of a single method -- I think the analogy to filesystems is pretty good. `ls` returns a list of contents, including files and folders; `mkdir` and `touch` are type-specific creation commands.

@junjun-zhang -- I agree that the DRS model is very similar to filesystems and to git, which is not surprising. The world has converged on a reasonable pattern for acessing possibly-nested content repositories, so it's good for us to stay close to that pattern. I do *not* think we should require DRS servers to be backed by a filesystem -- similar to HTTP servers, that's a fine implementation option, which many servers may choose, but other options (e.g. database-backed, multi-include labels vs. single-include folders) are also reasonable.

@junjun-zhang , re changing the `/access` method -- let's discuss that in #263.

sarpera:
@dglazer the issue with that is, it also makes it legit to have a DRS ID resolving into a metadata for a single file / blob that **also** has `contents`. Syntax-wise it passes as a legit response from the API standpoint. Another problem from legitimising this would be; do the urls in `access_methods` provide access to a single file or its `contents` in this case?

dglazer:
@sarpera , I agree that with this PR, servers could return a syntactically valid, but semantically invalid, object. (By returning `contents` for a single file-like blob.) My take, though, is that's a small price to pay for the benefits of method unification.

jaeddy:
I feel like we're overlooking a layer here: the job of DRS is *not* search. Some other service would presumably index some repository and provide a method by which to return DRS IDs for a particular query. These IDs are then passed to WES, but need to be mapped to the specific parameters in a job JSON/YAML file.

These parameters (for both CWL and WDL) have conventions for indicating whether they are a `File` or `Directory` (I'm not sure what the specific types are in WDL). Thus, when passing URL/URIs to a workflow engine, it makes sense to me to have some advanced indicator of whether a DRS ID represents a single object or a bundle — thus the engine can handle it appropriately.

On the other hand, if it's the job of WES to figure out whether a DRS ID represents an object or bundle, retrieve the bytes (or signed URLs) and *then* pass to the engine/executor, then having a unified endpoint is probably OK. However, given that I've yet to see any implementation of a "client" that a WES server could utilize, I'm also slightly concerned about this approach.

rishidev:
A 0 vote from AGHA

shaikh-rashid:
0 from CanDIG

openapi/data_repository_service.swagger.yaml
```diff
@@ -420,35 +347,26 @@ definitions:
...
         description: >-
           A list of full DRS identifier URI paths
-          that may be used obtain the Data Object or Data Bundle.
+          that may be used to obtain the object.
           These URIs may be external to this DRS instance.
         example:
           drs://example.com/ga4gh/drs/v1/objects/{object_id}
         items:
           type: string
-      type:
```
ddietterich: Why did this type information get removed? How are we to know what kind of object it is?

junjun-zhang: agree, should keep type

dglazer: I don't feel strongly here. I removed the type because callers can now just call `/objects/{object_id}` to get info about it including its type, which they need to do in all cases. So it felt redundant and like unneeded work for server developers to populate here. Given that, what do the two you think?

dglazer:
@jaeddy -- completely agree that DRS is not Search, and that a likely scenario is for a caller to call a Search API, get back a list of DRS IDs, and then pass those to WES. If WES needs the caller to pass in file/directory hints in the parameters, the caller is welcome to do so. I think that's completely consistent with DRS itself having a single method to get info about both blobs and bundes.

openapi/data_repository_service.swagger.yaml
ddietterich: I would prefer a restructuring so that the response was:
`[id, name, type (bundle or blob), bundle: BundleObject, object: BlobObject]`
and separate the two object structures. We can certainly share substructures of them, but not within this response.

That makes it obvious to the client which element to attend to based on the type. In code, it gives you one reference(pointer) to traverse. It also makes it clear how to extend the interface with additional types.

dglazer: Dan, are you suggesting that we embed the full object in the `contents` array, or am I misunderstanding? (We can talk on the WS call in a couple of minutes if easier.)

junjun-zhang: Had a closer look, didn't realize this `type` is under `BundleObject `. I thought `BundleObject` and *blob* object are merged, ie, there is no `BundleObject` any more. If that's the intention, I think it makes sense to keep `type` for the new unified `object`.

briandoconnor: I agree, I think this will be super useful when we have other object types in the future... like BigTable entries for example

dglazer: Notes from the WS call:
- Dan agrees we don't need to include the `type` in the `contents` array, but it could be useful -- he could take it or leave it
- Dan would prefer to restructure the entire response, to have a type at the top, and sub-structures below. But is okay either way.
- David pointed out that there the two sub-types are very similar (for blobs `access-methods` is required and `contents` isn't valid; for bundles `access-methods` is optional and `contents` is required)
- we want to rename `BundleObject` to `ContentsObject`

zflamig:
+1 from DCF

dvoet:
+1 from Terra

ddietterich:
+0 from Terra Data Repo

junjun-zhang:
+1 from ICGC-ARGO

jaeddy:
Are votes on this PR contingent on #272?

dglazer:
@jaeddy -- I suggest the votes be thought of independently. I expect some people will be +1 on one proposal and -0 on the other (meaning "I like X, but can live with Y"), or +0 on both (meaning "syntax shmyntax"). People are welcome to say -1/+1 (meaning "I hate X; we must do Y") if they explain their reasons.

jaeddy:
Thanks, @dglazer — was just a bit confused, as @ddietterich had both proposed changes and also voted +1. 🙂 

I realize that the priority is Drivers, but do you want votes from other Cloud API champions?

ddietterich:
Actually, I voted +0. My position is, I'd prefer that we do the bundle/object merger with the structure something like what I proposed in #272. But if we don't have time, I would prefer this PR approach over doing nothing.

sarpera:
-0 from Seven Bridges

xbrianh:
This is not congruous with the HCA data model, where bundles are flat namespaces (although file names within bundles may contain "/"), and entity URIs are not type ambiguous. This allows us to avoid the complexity of nested bundles while still providing file system structure to users. This seems desirable.

Also, while URI type hints may or may not be useful for machines, they are certainly useful for humans.

From the perspective of the HCA, we are -1.

mikebaumann:
I agree with @xbrianh, human usability of DRS identifiers is very important, and being able to identify if a given DRS URI is for a container/bundle or object is important to practical, human use. Conventional file system path names contain type hints, such as the presence of a file extension identifying a path is for a file, rather than a directory.
UUIDs and other generic identifiers don't provide this information, so having the DRS URI provide this fundamental fact is important for human (and computer) usability.

jaeddy:
@xbrianh, @mikebaumann what are the use cases for a human interacting with DRS IDs? My understanding was that users would typically identify objects/bundles through some indexing/search/query service (UI or otherwise), and that service would _return_ DRS IDs that get passed to a WES service or some other client that would handle access. In other words, consumption of DRS IDs would be predominantly managed by computers, not humans.

(not arguing one way or the other for object/bundle unification, but this seems to raise questions about the fundamental purpose of DRS)

dglazer:
@xbrianh, I think both of your comments (flat namespaces, non-ambiguous URIs) are unaffected by this PR, and therefore you'd be equally happy/unhappy with the proposed API whether or not this is merged. Is that right, or am I misunderstanding?

And if that is right, what API changes **would** you like to see? Is it:
a) allow "/" in `ContentsObject.name`
b) require DRS ids to have a prefix of "file:" or "dir:" (or equivalent) 
(Not sure how I'd vote on either of those; just trying to understand what would match your needs.)

dglazer:
Notes from a call with @xbrianh, @mikebaumann, and @briandoconnor 
- HCA has separate methods for readBundle and readFile
- HCA *could* define DRS IDs that include hints on whether they're a bundle or file (which would be opaque to others), and therefore could implement a single readObject method as defined here which wrapped the existing separate methods
- therefore, HCA is okay with this PR, and switches their vote to -0

And -- HCA thinks that, moving forward, future versions of DRS should consider some standard rules for ID and URI format.

dglazer:
We've now heard from all engaged driver projects, have a few +1's and a few +/- 0's. Merging.

PR #261, opened 2019-04-30
update URI convention (resolves #259)

### LINKED ISSUE #259: resolve URI convention
Splitting this off from #252  -- let's use this issue to discuss our desired URI convention, assuming that #258 (Bundle/Object unification) lands.

dglazer:


dglazer:
For easy reference: [HTML documentation](https://ga4gh.github.io/data-repository-service-schemas/preview/feature/issue-259-uri-convention/docs/)

docs/asciidoc/front_matter.adoc
```diff
@@ -15,6 +15,12 @@ Each implementation of DRS can choose its own id scheme, as long as it follows t
 * DRS v1 does NOT support semantics around multiple versions of an object. (For example, there’s no notion of “get latest version” or “list all versions”.) Individual implementation MAY choose an ID scheme that includes version hints.
 * DRS implementations MAY have more than one ID that maps to the same object.
 
+=== DRS URIs
+
+For convenience, including when passing content references to a WES server, we define a URI syntax for DRS-accessible content. Strings of the form `drs://<server>/<id>` mean _“you can fetch the content with DRS id `<id>` from the DRS server at `<server>` "_.
```
zflamig:
```
DRS IDs are URL-safe text strings made up of alphanumeric characters and any of [.-_/]
```

When we discussed in Hinxton we were saying that `/` wasn't valid so checking if a server existed would be trivial. However, that contradicts the other documents here referenced above. I propose we do `drs-<server>://<id>` or other such to make it trivial to separate in the future.

dglazer: Good catch @zflamig -- I was remembering that we prohibited slashes in `BundleObject.name`, and got that confused with IDs. But I think it's actually fine. Someone pointed out to me, `drs://<id>` isn't a valid URI, but `drs:<id>` is (which I think is supported by https://tools.ietf.org/html/rfc3986, but I gave up after 20 pages). So we're covered whenever we want to do a global ID scheme.

Personally, I still think slashes in IDs is bad practice, since life is easier if you never have to escape things. But I don't feel strongly, since escaping works.

mattions:
It seems there is a strong argument for making sure that the way we are going to define the Id with the `drs` in front is a proper URI.

For example, to connect with Phenopackets from the ClinoPheno stream, they have a`uri` specified there, and if `drs` is a proper URI, it could be slotted there without any need to have a dedicated namespace.

susheel:
Linked issue: #263

briandoconnor:
I think it would be helpful to address the following (even if we just say "this is something we'll decide on in the future but at least we're thinking about it :-) )

* what is the "right" way to do this from a URI perspective... is drs://server the URI-ish way of doing this?  Sorry if that's a stupid question but I'm seeing drs:<id> above based on David's comment.  I want to make sure our convention will not break other systems.
* we need to figure out issue #264 along with this I think.  It would be extremely useful to be able to make a /service-info request on any DRS service give the URI (e.g. given the server hostname).  That way we give a mechanism for a caller to know what version of DRS is supported and, therefore, how to turn DRS URIs into a successful https request (see my comment on the doc text in this PR)
* I think it would be great to get @zflamig and @dglazer to give some examples of how this would work right now with drs://<server>/<id> and how this might evolve to support GUIDs in the future.  Even if we don't document it in the spec, what's our gameplan for a future where GUID resolvers are super common and reliable?
* seems like the #263 issue needs to be resolved and make sure it's compatible with the above
* seems like #258 needs to land or be rejected before finalizing this

ddietterich:
Can someone explain (or point me to some other discussion explaining) the use case for "GUID resolvers"?

briandoconnor:
Oh, another thing to think about, secondary files.  Right now if I want to access a BAM and BAI it would like like:

drs://server/b869b801-ac5d-40f8-b27a-2de34913d79d for the BAM file
drs://server/35ecdbda-64e1-4f80-aaf7-212b2a3934df for the BAI file

That's well and good but if you have a URI for the BAM file you can't find the BAI file.

Something that's cool is if you could reference files via it's bundle.  For example, let's say the foo.bam and foo.bam.bai files are both in the bundle 50f43bdd-5e00-45f8-b658-a8f0b32bf65c we could write a URI like:

drs://server/50f43bdd-5e00-45f8-b658-a8f0b32bf65c/foo.bam
and
 drs://server/50f43bdd-5e00-45f8-b658-a8f0b32bf65c/foo.bam.bai

I can figure out that this is a bundle and, using the file name, could then figure out how to get the individual files very easily.  In this way, you can work with systems that will want to work with secondary files and be able to resolve them using a DRS URI.

@sarpera  what do you think?

sarpera:
@briandoconnor that assumes all files in a bundle are uniquely named. Current schema does not make that a requirement since bundle doesn't imply a folder-file structure.

Does that also mean, the client can make a GET request to a DRS server via `/50f43bdd-5e00-45f8-b658-a8f0b32bf65c/foo.bam.bai`? Practically, `50f43bdd-5e00-45f8-b658-a8f0b32bf65c/foo.bam.bai` is an alias/one of the aliases to 'foo.bam.bai'? 
I also fear having the file name in the URI might expose some information that may be contained in the file name. Some protected/private files might contain parts of patient/case/study information that are otherwise protected with authN/Z mechanisms in DRS.

rishidev:
A 0 vote from AGHA

shaikh-rashid:
0 from CanDIG

briandoconnor:
@sarpera based on your comments it sounds like the interaction of secondary files with the URI scheme should be handled in another PR.

@dglazer I'd like to finish my review and approve but I'm waiting for @zflamig and others to comment on my questions above.  I want to make sure our drivers are happy and I also think those linked tickets need to be addressed before this one gets merged in.

zflamig:
-1 from DCF

dglazer:
@zflamig -- thanks for flagging that you still have unresolved concerns. Can you say more about the specific v1-blocking issues? I assume it's the question you raise above about being able to parse out the server name from the possibly-slash-containing id -- is that right, or are there others? 

If that's it, my first thought is that standard URI rules should work fine, as they do for http: URIs where the page ID contains a slash. But I might be missing something.

dglazer:
@briandoconnor , re "changes requested" -- my take on the questions you raise:
- yes, we need #258 to land, so we know we don't need b/o qualifiers -- that's now looking likely to happen soon as the last few votes come in
- yes, we should resolve #264, because of the v1/v2 question you raise. I know you're pushing on that; worst case we're okay as is if we drop /service-info support in v1, since callers can assume /v1 until we introduce explicit version support in v2.
- I agree we should be be prepared for a potential future upgrade to support a central universal URI-resolver. I think that's already the case, either using the `drs:<id>` model (with no double-slash), or using a new prefix (e.g. `drs-guid:<id>`).
- it will be good to resolve #263, but I think this PR already works with slashes in IDs, and life is even easier if we decide to prohibit them, so that shouldn't be blocking
- I believe we agreed to defer secondary file support

ddietterich:
+1 from Terra Data Repo

junjun-zhang:
+1 from ICGC-ARGO

sarpera:
+0 from Seven Bridges

PR #271, opened 2019-05-20
Document the use of reserved & unreserved characters in DRS IDs

### LINKED ISSUE #263: Should we allow slashes in DRS IDs?
Following our last meeting, there was a discussion if we should explicitly disallow forward-slashes `/` in a DRS `id`. Zac (NIH Data Commons) uses dataguids and/or ark ids as DRS ids which have forward-slashes in them.

The current documentation does allow forward-slash. See [front-matter.adocL#13](https://github.com/ga4gh/data-repository-service-schemas/blob/master/docs/asciidoc/front_matter.adoc#L13)
> DRS IDs are URL-safe text strings made up of alphanumeric characters and any of [.-_/]

Error Conditions:

1. GET `drs://<server>/<foo/access>` (where `<foo/access>` is the DRS id)
2. GET `drs://<server>/<foo/access/1234>` (where `<foo/access>` is the DRS id)

The above requests will not return the expected `Object` definition, but an `Error` definition as the DRS server expects an `access_id` after `/access` as per the [current spec](https://github.com/ga4gh/data-repository-service-schemas/blob/cd0186fa9b7763bdb15175e6615ff4a42c1b85c3/openapi/data_repository_service.swagger.yaml#L117)

Possible Solutions:

- Disallow `forward-slashes` in a DRS `id`; OR
- Restructure the AccessMethod API:
  From: `GET /objects/{object_id}/access/{access_id}`
  To: `GET /access/{access_id}`

susheel:
Doc Fixes:
* DRS IDs are strings made up of uppercase and lowercase letters, decimal digits, hypen, period, underscore and tilde [A-Za-z0-9.-_~]. See [IETF RFC 3986 § 2.3](https://tools.ietf.org/html/rfc3986#section-2.3).
* ~~All RFC 3986 § 3986 reserved characters including whitespace in DRS IDs must be url-encoded [!*'();:@&=+$,/?#[] ]. See [IETF RFC 3986 § 2.2](https://tools.ietf.org/html/rfc3986#section-2.2).~~
* All reserved characters including whitespace [!*'();:%@&=+$,/?#[] ] in DRS IDs must be may be custom-encoded only using unreserved characters [A-Za-z0-9.-_~]. Note: DO NOT expect this custom encoding to be supported by other DRS implementors.

Resolves #263

### DIFF
--- a/docs/asciidoc/front_matter.adoc
+++ b/docs/asciidoc/front_matter.adoc
@@ -11,7 +11,6 @@ The primary functionality of DRS is to map a logical ID to a means for physicall
 Each implementation of DRS can choose its own id scheme, as long as it follows these guidelines:
 
 * DRS IDs are strings made up of uppercase and lowercase letters, decimal digits, hypen, period, underscore and tilde [A-Za-z0-9.-_~]. See https://tools.ietf.org/html/rfc3986#section-2.3[RFC 3986 § 2.3].
-* All RFC 3986 § 3986 reserved characters including whitespace in DRS IDs must be url-encoded [!*'();:@&=+$,/?#[] ]. See https://tools.ietf.org/html/rfc3986#section-2.2[RFC 3986 § 2.2].
 * One DRS ID MUST always return the same object data (or, in the case of a collection, the same set of objects). This constraint aids with reproducibility.
 * DRS v1 does NOT support semantics around multiple versions of an object. (For example, there’s no notion of “get latest version” or “list all versions”.) Individual implementation MAY choose an ID scheme that includes version hints.
 * DRS implementations MAY have more than one ID that maps to the same object.
--- a/docs/asciidoc/front_matter.adoc
+++ b/docs/asciidoc/front_matter.adoc
@@ -10,7 +10,8 @@ The primary functionality of DRS is to map a logical ID to a means for physicall
 
 Each implementation of DRS can choose its own id scheme, as long as it follows these guidelines:
 
-* DRS IDs are URL-safe text strings made up of alphanumeric characters and any of [.-_/]
+* DRS IDs are strings made up of uppercase and lowercase letters, decimal digits, hypen, period, underscore and tilde [A-Za-z0-9.-_~]. See https://tools.ietf.org/html/rfc3986#section-2.3[RFC 3986 § 2.3].
+* All RFC 3986 § 3986 reserved characters including whitespace in DRS IDs must be url-encoded [!*'();:@&=+$,/?#[] ]. See https://tools.ietf.org/html/rfc3986#section-2.2[RFC 3986 § 2.2].
 * One DRS ID MUST always return the same object data (or, in the case of a collection, the same set of objects). This constraint aids with reproducibility.
 * DRS v1 does NOT support semantics around multiple versions of an object. (For example, there’s no notion of “get latest version” or “list all versions”.) Individual implementation MAY choose an ID scheme that includes version hints.
 * DRS implementations MAY have more than one ID that maps to the same object.
--- a/docs/asciidoc/front_matter.adoc
+++ b/docs/asciidoc/front_matter.adoc
@@ -11,6 +11,7 @@ The primary functionality of DRS is to map a logical ID to a means for physicall
 Each implementation of DRS can choose its own id scheme, as long as it follows these guidelines:
 
 * DRS IDs are strings made up of uppercase and lowercase letters, decimal digits, hypen, period, underscore and tilde [A-Za-z0-9.-_~]. See https://tools.ietf.org/html/rfc3986#section-2.3[RFC 3986 § 2.3].
+* All reserved characters including whitespace [!*'();:%@&=+$,/?#[] ] in DRS IDs must be may be custom encoded only using unreserved characters [A-Za-z0-9.-_~]. Note: DO NOT expect this custom encoding to be supported by other DRS implementors.
 * One DRS ID MUST always return the same object data (or, in the case of a collection, the same set of objects). This constraint aids with reproducibility.
 * DRS v1 does NOT support semantics around multiple versions of an object. (For example, there’s no notion of “get latest version” or “list all versions”.) Individual implementation MAY choose an ID scheme that includes version hints.
 * DRS implementations MAY have more than one ID that maps to the same object.
--- a/docs/asciidoc/front_matter.adoc
+++ b/docs/asciidoc/front_matter.adoc
@@ -11,7 +11,7 @@ The primary functionality of DRS is to map a logical ID to a means for physicall
 Each implementation of DRS can choose its own id scheme, as long as it follows these guidelines:
 
 * DRS IDs are strings made up of uppercase and lowercase letters, decimal digits, hypen, period, underscore and tilde [A-Za-z0-9.-_~]. See https://tools.ietf.org/html/rfc3986#section-2.3[RFC 3986 § 2.3].
-* All reserved characters including whitespace [!*'();:%@&=+$,/?#[] ] in DRS IDs must be may be custom encoded only using unreserved characters [A-Za-z0-9.-_~]. Note: DO NOT expect this custom encoding to be supported by other DRS implementors.
+* Note to server implementors: internal IDs can contain other characters, but they MUST be encoded into valid DRS IDs whenever exposed by the API.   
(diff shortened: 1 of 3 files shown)


docs/asciidoc/front_matter.adoc
```diff
@@ -10,7 +10,8 @@ The primary functionality of DRS is to map a logical ID to a means for physicall
 
 Each implementation of DRS can choose its own id scheme, as long as it follows these guidelines:
 
-* DRS IDs are URL-safe text strings made up of alphanumeric characters and any of [.-_/]
+* DRS IDs are strings made up of uppercase and lowercase letters, decimal digits, hypen, period, underscore and tilde [A-Za-z0-9.-_~]. See https://tools.ietf.org/html/rfc3986#section-2.3[RFC 3986 § 2.3].
+* All RFC 3986 § 3986 reserved characters including whitespace in DRS IDs must be url-encoded [!*'();:@&=+$,/?#[] ]. See https://tools.ietf.org/html/rfc3986#section-2.2[RFC 3986 § 2.2].
```
briandoconnor: This seems innocuous to me, so if a given system needs "/" in the ID, for example, it can still be supported.  But when using DRS URIs/URLs we can count on "/" as a delimiter.

briandoconnor:
@susheel I think we should do a similar thing for the bundle/file name fields.  Right now the doc say you can't have a "/" in the file name field but systems that want to support DRS (like the HCA data store) actually do use "/" in file names.  In that project we treat a bundle like an S3 bucket and all the directory structure inside a bundle is expressed using "/" in the file names.  We could support your approach of using bundles as folders and the HCA approach of using "/" in the name by basically saying to URL encode the name field just as you did in the ID PR here.

@xbrianh and @mikebaumann is this correct?

dglazer:
@susheel , after today's Cloud WS call I realized that people could interpret this PR in two different ways. Picking them apart:

(1) **strict:** DRS IDs can only contain the 66 characters in RFC 3986
- the only difference from the v0.1 spec is that we disallow slashes in DRS IDs
- server developers are welcome to use arbitrary internal IDs, and use any encoding they want to map them to/from DRS IDs
- this approach makes life easy for client developers and us as API designers, since DRS IDs are opaque, URL-safe, and fit into easily-parseable URLs
- we don't need the second line of your suggested edit, since there's no URL-encoding needed. (Plus it wouldn't be possible, since % characters aren't allowed.)

(2) **loose:** DRS IDs can contain any character, as long as they're URL-encoded when passed around
- server developers can use any DRS ID scheme they want, but have to add encode/decode logic on input and output
- client developers also have to add encode/decode logic when receiving and sending DRS IDs
- we don't need the first line of your suggested edit, since there aren't actually any restrictions on the IDs, only on how they're passed around

If that's right, I think you intended (1) strict, and I agree. I suggest just deleting the second line of your edit, and then I'd be +1.

@briandoconnor , re the `name` fields -- thanks for creating #276; let's discuss there.

susheel:
@dglazer As per your comment above. I have made it stricter.

dglazer:
Thanks @susheel -- this change now LGTM. Somehow it looks like an unrelated (and incorrect) edit to README.md snuck in -- can you undo that, and then I'll be happy to approve?

@briandoconnor -- please re-review, since I'm not sure if you preferred the strict or loose interpretation, and the initial PR was halfway between.

mattions:
+1 from CGC and TOPMed from SB point of view.

susheel:
+0 ELIXIR

junjun-zhang:
+1 from ICGC-ARGO

Agree with @susheel, being able to reference a blob by `/<bundle_id>/<object_id>` would be great, although `<object_id>` is globally unique and can be addressed directly.

Going through `bundle` makes it easier to support secondary files which are kept under the same `bundle` as the main file.

briandoconnor:
@susheel and @junjun-zhang I left some similar comments in the PR for URI about wanting to do a similar thing with `/bundle_id/file_name`... I think it's a very appealing way of handling secondary files referenced via the bundle but in such a way that allows you to refer to secondary files without worrying about different object_ids for each.  Something to discuss there I think: PR #261 

For the loose vs. strict debate I think it's short sighted to go with the strict approach.  While it works for drivers I represent, because we happen to use UUIDs for our IDs, I'm really afraid it paints us into a corner here.  If we allow people to URL encode in the spec, then they can put whatever they want in the ID and it won't effect the API in any way.  They can internally encode/decode to work with their preferred ID and the calling DRS client can just work with the encoded IDs directly... there's no need for clients to encode/decode IDs since the encoded ID is just the DRS ID they work with.

If we go strict we're saying systems that don't conform to this need to maintain two sets of IDs... one they expose via DRS and one they use internally or elsewhere in their stack to refer to data.  I think it's really easy for us to say "oh they can just maintain this mapping" but by allowing implementers to use whatever ID scheme they want we lower the bar to entry for implementers of DRS.  And I think this is a good thing, we need to make it easy for people to implement DRS so we can get a lot of content available in that system.  This is a trivial thing but a small choice like this could make or break a project from adopting this.

My take, URL encoding doesn't complicate any implementation that doesn't use it, it doesn't complicate the client since it's just an ID (encoded or not), but it does allow us to support projects with a wide variety of ID schemes and we all know how concerned people get about their IDs :-)

dglazer:
@susheel, @junjun-zhang, re secondary files, and possibly using `/bundle_id/file_name` to handle them -- seems like a promising approach; let's keep that in mind for when we're ready to add secondary file support later. (Note that if we do want to use that approach later, it's a further argument for this PR as it stands, which prohibits slashes in DRS IDs.)

@briandoconnor , re loose vs. strict -- I'm not sure i follow. When you say you support **loose**, that would mean you do **not** approve of this PR as it stands (even though it works for all of our drivers), and want to require all clients to use URL encoding/decoding on all calls to all servers. On the other hand, you also say
> My take, URL encoding doesn't complicate any implementation that doesn't use it, it doesn't complicate the client since it's just an ID (encoded or not)

I think that's actually a argument for **strict**, since you're saying it's up to servers that care about unsafe ID characters to encode/decode them before they're seen by the outside world. Let's talk on today's call to clarify.

ddietterich:
If we are going to have endpoints like `/objects/{object_id}/access/{access_id}` then there is no question: the `{object_id}` cannot have a slash in it or the endpoint is ambiguous and can't be reliably parsed.

We could allow slash if we change DRS to remove object_id from any endpoint path and instead put it in the body or in a query parameter, but that seems clunky and not RESTful to me.

If you want to use slash in your non-DRS interfaces and inside your implementation, go right ahead, but you have to do something to make it URI-legal when used in DRS endpoints. As suggested above, an obvious implementation for such an implementation is to say, "my DRS id is the URL encoding of my implementation id." But that is not really a DRS issue. You could choose many implementations for id mapping.

briandoconnor:
Yep, @ddietterich, I think you expressed what I was thinking much better than I.  I'd like to give DRS implementers the option to say "my DRS id is the URL encoding of my implementation id".  However if we remove the sentence that @susheel had originally added, then the regex of allowed characters is only `[A-Za-z0-9.-_~]`.  Hence, if I'm using data GUIDs as my IDs, I can't encode `dg.4503/00e6cfa9-a183-42f6-bb44-b70347106bbe` as `dg.4503%2F00e6cfa9-a183-42f6-bb44-b70347106bbe` since the `%` character is not allowed.  To work around this I could maintain a mapping of external to internal IDs (work for implementers) or come up with my own encoding scheme that uses the allowed characters (but why, we have a perfectly good encoding scheme with URL-encoding that will not break parsing our URLs like `/objects/dg.4503%2F00e6cfa9-a183-42f6-bb44-b70347106bbe/access/1`)

@dglazer hopefully this clarifies that I'm going for your loose option here because I think it will ultimately serve us better to allow implementers to use URL encoding to support the use of data GUIDs, arks, etc as IDs for objects in DRS without breaking our endpoints.

Let's discuss on today's call

susheel:
Added additional text below:
> * All reserved characters including whitespace [!*'();:%@&=+$,/?#[] ] in DRS IDs must be may be custom-encoded only using unreserved characters [A-Za-z0-9.-_~]. Note: DO NOT expect this custom encoding to be supported by other DRS implementors.

ddietterich:
+1 Terra Data Repo

briandoconnor:
+1 for HCA

PR #272, opened 2019-05-20
Feature/issue 252 object bundle merge alt

ddietterich:
As we discussed in today's meeting, here is a draft pull request that restructures the object response. I think this has a few things going for it:
- It makes a cleaner separation at the top-level with only one choice for clients: either follow bundle or follow blob.
- It allows independent specification of what properties are required for blob and bundle
- It provides a structure for extending to other types, if we choose to do that.

### DIFF
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -154,7 +154,7 @@ definitions:
           sha512
   Object:
     type: object
-    required: ['id', 'size', 'created', 'checksums']
+    required: ['id']
     properties:
       id:
         type: string
@@ -164,12 +164,23 @@ definitions:
         type: string
         description: |-
           A string that can be used to name an `Object`.
-      size:
-        type: integer
-        format: int64
+      type:
+        type: string
+        enum:
+        - blob
+        - bundle
+      blob:
+        $ref: '#/definitions/BlobDescription'
         description: |-
-          For blobs, the blob size in bytes.
-          For bundles, the cumulative size, in bytes, of items in the `contents` field.
+          If `type` is blob, then this property must be present and the `bundle` property must be absent or null
+      bundle:
+        $ref: '#/definitions/BundleDescription'
+        description: |-
+          If `type` is bundle, then this property must be present and the `blob` property must be absent or null
+  CommonDescription:
+    type: object
+    required: ['created']
+    properties:
       created:
         type: string
         format: date-time
@@ -187,6 +198,31 @@ definitions:
           A string representing a version.
 
           (Some systems may use checksum, a RFC3339 timestamp, or an incrementing version number.)
+      description:
+        type: string
+        description: |-
+          A human readable description of the `Object`.
+      aliases:
+        type: array
+        items:
+          type: string
+        description: >-
+          A list of strings that can be used to find other metadata 
+          about this `Object` from external metadata sources. These
+          aliases can be used to represent secondary
+          accession numbers or external GUIDs.
+  BlobDescription:
+    type: object
+    required: ['common', 'size', 'checksums', 'access_methods']
+    properties:
+      common:
+        $ref: '#/definitions/CommonDescription'
+      size:
+        type: integer
+        format: int64
+        description: |-
+          For blobs, the blob size in bytes.
+          For bundles, the cumulative size, in bytes, of items in the `contents` field.
       mime_type:
         type: string
         description: |-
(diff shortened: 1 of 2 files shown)


openapi/data_repository_service.swagger.yaml
```diff
@@ -192,144 +152,156 @@ definitions:
...
+        description: |-
+          An identifier unique to this `Object`.
       name:
         type: string
-        description: >-
-          A string that can be used to name a Data Bundle.
-      size:
+        description: |-
+          A string that can be used to name an `Object`.
+      type:
```
jaeddy: `type` is a keyword in Swagger and JSON Schema (and `id` is a keyword in JSON Schema). Some of our other APIs might be guilty of this as well, but I wonder if we should try harder to avoid keywords in our property names (e.g., could use `object_id` and `object_type`).

Will open a separate issue for discussion.

openapi/data_repository_service.swagger.yaml
```diff
@@ -192,144 +152,156 @@ definitions:
...
+        - bundle
+      blob:
+        $ref: '#/definitions/BlobDescription'
+        description: |-
+          If `type` is blob, then this property must be present and the `bundle` property must be absent or null
+      bundle:
+        $ref: '#/definitions/BundleDescription'
+        description: |-
+          If `type` is bundle, then this property must be present and the `blob` property must be absent or null
+  CommonDescription:
```
jaeddy: Why `-Description`? This again implies something specific in the world of Swagger/OpenAPI. What's being specified here is the 'schema' for an object (in the Swagger sense of the word 'object') — `CommonObject` seems more appropriate to me.

openapi/data_repository_service.swagger.yaml
```diff
@@ -192,144 +152,156 @@ definitions:
...
-              = md5( concat( sort( 72794b6d30bc86d92e40a1aa65c880b8, 5e089d29a18954e68a78ee6a3c6edabd ) ) )
-              = md5( concat( 5e089d29a18954e68a78ee6a3c6edabd, 72794b6d30bc86d92e40a1aa65c880b8 ) )
-              = md5( 5e089d29a18954e68a78ee6a3c6edabd72794b6d30bc86d92e40a1aa65c880b8 )
-              = f7a29a0422e7d870b10839ad6c985079
-        items:
-          $ref: '#/definitions/Checksum'
+          A string representing a version.
+
+          (Some systems may use checksum, a RFC3339 timestamp, or an incrementing version number.)
       description:
```
jaeddy: I might retract (or revise) my comment above regarding keywords... I have no objections to a `description` property that also has a Swagger/OpenAPI/JSON Schema `description`.

openapi/data_repository_service.swagger.yaml
```diff
@@ -192,144 +152,156 @@ definitions:
...
         description: >-
-          The list of Data Objects and Data Bundles contained by this Data Bundle.
-        items:
-          $ref: '#/definitions/BundleObject'
-  Object:
+          A list of strings that can be used to find other metadata 
+          about this `Object` from external metadata sources. These
+          aliases can be used to represent secondary
+          accession numbers or external GUIDs.
+  BlobDescription:
```
jaeddy: Would change to `BlobObject` or just `Blob`.

openapi/data_repository_service.swagger.yaml
jaeddy: I see that this is used more as a `property` than a 'base' model for inheritance, etc. `CommonProperties` might make more sense.

openapi/data_repository_service.swagger.yaml
```diff
@@ -192,144 +152,156 @@ definitions:
...
         type: array
         minItems: 1
         items:
           $ref: '#/definitions/AccessMethod'
         description: |-
-          The list of access methods that can be used to fetch the Data Object.
-      description:
-        type: string
+          The list of access methods that can be used to fetch the `Object`.
+  BundleDescription:
```
jaeddy: Would change to `BundleObject` or just `Bundle`.

openapi/data_repository_service.swagger.yaml
```diff
@@ -414,41 +386,32 @@ definitions:
       license:
         type: object
         description: License information for the exposed API
-  BundleObject:
+  ContentsObject:
```
jaeddy: So, a `ContentsObject` is basically a 'reference' to a blob or bundle? Might be worth stating so in the description.

dglazer:
Useful diff link: https://github.com/ga4gh/data-repository-service-schemas/compare/feature/issue-252-object-bundle-merge...ddietterich:feature/issue-252-object-bundle-merge-alt

(still thinking about pros and cons of this approach)

dglazer:
_tl;dr: I'm -0 on this approach for v1. I like parts of it, but all-in-all don't find it clearly better, and think the tradeoffs will be clearer when we add more object types down the road._

Pros of this approach:
- much cleaner top-level `Object` definition (id, name, type, type-specific description)
- easier to extend to future object types (but see the first Con below)
- easier for a client who only cares about one object type at a time to understand the documentation

Cons of this approach:
- as we add more object types, we may end up with properties that are shared between blobs and bundles, but not e.g. with tables. Then we'd either need more repetition of properties, or a Venn diagram of e.g. `CommonBlobBundleDescription` objects
- longer overall schema description, with some repetition across object types
- nested schema (not sure if that applies to wire format and parsing code, or just to the yaml)

Minor comments on the details (all of which move towards more common stuff, and therefore fewer Pros):
- I think `size` and `checksums` are identical between blobs and bundles (except for clarifying comments on how they're computed), so would move them into `CommonDescription`
- the only difference between blob and bundle `access_methods` is whether it's required or optional; I lean towards making that Common
- I think bundles need an optional `mime_type`, which would be used whenever using `access_methods` to download them as a single entity'
- not sure what belongs in `Object` vs. in `CommonDescription`

PR #276, opened 2019-06-03
Document the use of reserved & unreserved characters in DRS name fields (resolves #275)

### LINKED ISSUE #275: Should we allow slashes in DRS filenames?
Both Bundle and BundleObject have a name field.  Currently, the latter requires no slashes however we have DRS implementations out there that use slashes in the file name to replicate directory structure within a give bundle (e.g. HCA).  Also, another concern is there's no requirement that bundleobject names be unique within a bundle and if you're trying to download an entire bundle filled with identically named files this is a problem.

A proposal:

* for systems that want to use "/" in file names we allow it however we encode it
* like IDs, anything beyond [A-Za-z0-9.-_]. See https://tools.ietf.org/html/rfc3986#section-2.3[RFC 3986 § 2.3]. will need to be encoded (I removed ~ from this)
* encode [!*'();:@&=+$,/?#[] ]. See https://tools.ietf.org/html/rfc3986#section-2.2[RFC 3986 § 2.2].
* file names must be unique in a bundle

A nice side effect is, if we ever want to use file names within URIs they will be safely encoded.

I'm taking the above from #263 

 @xbrianh, @mikebaumann what do you think since this is being driven by the HCA Data Store's use case

briandoconnor:
See #275 for more details.

### DIFF
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -164,6 +164,8 @@ definitions:
         type: string
         description: |-
           A string that can be used to name an `Object`.
+          This string is made up of uppercase and lowercase letters, decimal digits, hypen, period, and underscore [A-Za-z0-9.-_]. See https://tools.ietf.org/html/rfc3986#section-2.3[RFC 3986 § 2.3].
+          All RFC 3986 § 3986 reserved characters including whitespace in DRS IDs must be url-encoded [!*'();:@&=+$,/?#[] ]. See https://tools.ietf.org/html/rfc3986#section-2.2[RFC 3986 § 2.2].
       size:
         type: integer
         format: int64
@@ -350,7 +352,9 @@ definitions:
           A name declared by the bundle author that must be
           used when materialising this object, 
           overriding any name directly associated with the object itself.
-          This string MUST NOT contain any slashes.
+          If this is a content object inside of a bundle then the name must be unique within that bundle. 
+          This string is made up of uppercase and lowercase letters, decimal digits, hypen, period, and underscore [A-Za-z0-9.-_]. See https://tools.ietf.org/html/rfc3986#section-2.3[RFC 3986 § 2.3].
+          All RFC 3986 § 3986 reserved characters including whitespace in DRS IDs must be url-encoded [!*'();:@&=+$,/?#[] ]. See https://tools.ietf.org/html/rfc3986#section-2.2[RFC 3986 § 2.2].
       id:
         type: string
         description: >-
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -165,7 +165,7 @@ definitions:
         description: |-
           A string that can be used to name an `Object`.
           This string is made up of uppercase and lowercase letters, decimal digits, hypen, period, and underscore [A-Za-z0-9.-_]. See https://tools.ietf.org/html/rfc3986#section-2.3[RFC 3986 § 2.3].
-          All RFC 3986 § 3986 reserved characters including whitespace in DRS IDs must be url-encoded [!*'();:@&=+$,/?#[] ]. See https://tools.ietf.org/html/rfc3986#section-2.2[RFC 3986 § 2.2].
+          All RFC 3986 § 3986 reserved characters including whitespace in the object name must be url-encoded [!*'();:@&=+$,/?#[] ]. See https://tools.ietf.org/html/rfc3986#section-2.2[RFC 3986 § 2.2].
       size:
         type: integer
         format: int64
@@ -352,9 +352,9 @@ definitions:
           A name declared by the bundle author that must be
           used when materialising this object, 
           overriding any name directly associated with the object itself.
-          If this is a content object inside of a bundle then the name must be unique within that bundle. 
+          The name must be unique with the containing bundle. 
           This string is made up of uppercase and lowercase letters, decimal digits, hypen, period, and underscore [A-Za-z0-9.-_]. See https://tools.ietf.org/html/rfc3986#section-2.3[RFC 3986 § 2.3].
-          All RFC 3986 § 3986 reserved characters including whitespace in DRS IDs must be url-encoded [!*'();:@&=+$,/?#[] ]. See https://tools.ietf.org/html/rfc3986#section-2.2[RFC 3986 § 2.2].
+          All RFC 3986 § 3986 reserved characters including whitespace in the object name must be url-encoded [!*'();:@&=+$,/?#[] ]. See https://tools.ietf.org/html/rfc3986#section-2.2[RFC 3986 § 2.2].
       id:
         type: string
         description: >-
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -164,8 +164,7 @@ definitions:
         type: string
         description: |-
           A string that can be used to name an `Object`.
-          This string is made up of uppercase and lowercase letters, decimal digits, hypen, period, and underscore [A-Za-z0-9.-_]. See https://tools.ietf.org/html/rfc3986#section-2.3[RFC 3986 § 2.3].
-          All RFC 3986 § 3986 reserved characters including whitespace in the object name must be url-encoded [!*'();:@&=+$,/?#[] ]. See https://tools.ietf.org/html/rfc3986#section-2.2[RFC 3986 § 2.2].
+          This string is made up of uppercase and lowercase letters, decimal digits, hypen, period, and underscore [A-Za-z0-9.-_]. See http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_282[portable filenames].
       size:
         type: integer
         format: int64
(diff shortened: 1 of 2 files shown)


openapi/data_repository_service.swagger.yaml
```diff
@@ -164,6 +164,8 @@ definitions:
         type: string
         description: |-
           A string that can be used to name an `Object`.
+          This string is made up of uppercase and lowercase letters, decimal digits, hypen, period, and underscore [A-Za-z0-9.-_]. See https://tools.ietf.org/html/rfc3986#section-2.3[RFC 3986 § 2.3].
+          All RFC 3986 § 3986 reserved characters including whitespace in DRS IDs must be url-encoded [!*'();:@&=+$,/?#[] ]. See https://tools.ietf.org/html/rfc3986#section-2.2[RFC 3986 § 2.2].
```
dglazer: I suggest deleting the sentence on URL-encoding  -- these aren't IDs, and they don't get passed in URLs so don't need to be encoded.

openapi/data_repository_service.swagger.yaml
```diff
@@ -350,7 +352,9 @@ definitions:
           A name declared by the bundle author that must be
           used when materialising this object, 
           overriding any name directly associated with the object itself.
-          This string MUST NOT contain any slashes.
+          If this is a content object inside of a bundle then the name must be unique within that bundle. 
+          This string is made up of uppercase and lowercase letters, decimal digits, hypen, period, and underscore [A-Za-z0-9.-_]. See https://tools.ietf.org/html/rfc3986#section-2.3[RFC 3986 § 2.3].
+          All RFC 3986 § 3986 reserved characters including whitespace in DRS IDs must be url-encoded [!*'();:@&=+$,/?#[] ]. See https://tools.ietf.org/html/rfc3986#section-2.2[RFC 3986 § 2.2].
```
dglazer: I suggest deleting the sentence on URL-encoding  -- these aren't IDs, and they don't get passed in URLs so don't need to be encoded.

openapi/data_repository_service.swagger.yaml
```diff
@@ -350,7 +352,9 @@ definitions:
           A name declared by the bundle author that must be
           used when materialising this object, 
           overriding any name directly associated with the object itself.
-          This string MUST NOT contain any slashes.
+          If this is a content object inside of a bundle then the name must be unique within that bundle.
```
dglazer: We *know* this is a content object inside a bundle -- maybe simplify to "The name must be unique with the containing bundle."

openapi/data_repository_service.swagger.yaml
```diff
@@ -164,6 +164,8 @@ definitions:
         type: string
         description: |-
           A string that can be used to name an `Object`.
+          This string is made up of uppercase and lowercase letters, decimal digits, hypen, period, and underscore [A-Za-z0-9.-_]. See https://tools.ietf.org/html/rfc3986#section-2.3[RFC 3986 § 2.3].
```
dglazer: Is there a reason that we want names to use the same characters as URIs? Maybe reference [POSIX filenames](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_282) instead?

openapi/data_repository_service.swagger.yaml
```diff
@@ -350,7 +352,9 @@ definitions:
           A name declared by the bundle author that must be
           used when materialising this object, 
           overriding any name directly associated with the object itself.
-          This string MUST NOT contain any slashes.
+          If this is a content object inside of a bundle then the name must be unique within that bundle. 
+          This string is made up of uppercase and lowercase letters, decimal digits, hypen, period, and underscore [A-Za-z0-9.-_]. See https://tools.ietf.org/html/rfc3986#section-2.3[RFC 3986 § 2.3].
```
dglazer: Is there a reason that we want names to match URIs? Maybe cite [POSIX filenames](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_282) instead? (I think it's the same characters, but makes more sense as a reference.)

dglazer:
I added several small comments inline.

Overall, I think being clearer about valid `name` characters makes sense. The one potentially controversial part is whether to allow slashes, which isn't changed here -- they were prohibited before, and still are now. We discussed that on the Cloud WS call just now, and the general feeling was that continuing to prohibit slashes makes sense, since you can't create files with slashes in their names, so it's not clear what a client would do if told to instantiate a local copy of an object with slashes in its name (e.g. to pass to WES).

@xbrianh @mikebaumann -- I'd like to hear HCA's thoughts, since IIUC you currently use slashes in names. How do you expect clients to handle that? Is the intent that, if a client sees a `ContentsObject.name` with a slash, it should create subdirectories as needed?

susheel:
+1 ELIXIR

junjun-zhang:
+ 1 from ICGC-ARGO for only allowing `[A-Za-z0-9.-_]` in `name`

openapi/data_repository_service.swagger.yaml
briandoconnor: yep I like that

openapi/data_repository_service.swagger.yaml
briandoconnor: I changed this to say name, the reference to ID was a typo.

openapi/data_repository_service.swagger.yaml
briandoconnor: Let's talk about this tomorrow so I can be sure I understand what you mean

briandoconnor:
On the HCA side our client uses the "/" character to materialize directories and files correctly. 

This PR is a compromise, by allowing for URL-encoding it means we can continue to use "/" via encoding in the file name but not interfere with other implementations that want a prohibition of this character in the name.  Both approaches can coexist.  A client can naively download a file from an HCA bundle using the encoded name and that will just work as a file name.  Likewise, same goes for a system that only uses POSIX characters in their name.  

The alternatively of strictly prohibiting this mean that the HCA implementation will need to break it's bundles down into a DRS representation that creates a bunch of fake bundles that replicate the structure and carefully map those to the true bundles that allow directory structure within them.  I think this is a huge ask for this driver project and means we're asking them to adapt to DRS and not the other way around.

A third option is to just allow "/" in the names and say if a client sees that they are supposed to create directory structures as needed when downloading a bundle.

Of the three options:

1. no slashes
2. slashes but encoded (along with other forbidden characters)
3. slashes allowed and interpreted by a client

I think HCA would like 2 or 3 but I'll let Michael and Brian H weigh in

dglazer:
I understand option 1 (status quo -- no slashes) and option 3 (slashes tell clients to make directories). I don't understand option 2 (encoded slashes), since if clients ignore the encoding it's equivalent to option 1, and doesn't result in the local directory structure I thought HCA needed.

junjun-zhang:
with object/bundle unification PR (#258) landed, bundles can contain sub-bundles / blobs. This should enable implementers (if so wish) to mimic folder structure.

What I have in mind is how git works, files are blob objects, folders are tree objects. IDs for these objects are hash codes (no slashes in them). Names for those objects are the actual names of the files/folders, as such no slashes as well.

My point is not about any particular implementation, it's more of about banning slashes in `ID` and `name` does not prevent an implementer from supporting arbitrary folder structure.

openapi/data_repository_service.swagger.yaml
briandoconnor: I found this really helpful reference here: https://en.wikipedia.org/wiki/Filename#Comparison_of_filename_limitations

briandoconnor:
@junjun-zhang there's nothing that prevents you for implementing this way, the spec supports nested bundles.  However, in HCA we have bundles that contain millions of files, all with directory structures.  So if we support "/" in names we can have a 1:1 mapping of bundle to our bundle in HCA to a bundle in DRS, nice!

If we aren't allowed to have "/" in the file names then we are going to need to make thousands of bundles to reconstruct a bundle currently in the HCA system.  That's OK, but a lot of work for developers on my team to write that code.  Furthermore, when a calling client tries to discover the structure of a bundle in your approach, they will need to make thousands of GET requests to fully explore the structure of a given bundle by following the nesting.  This will cause performance problems for us.

So, at the end of the day HCA has two choices 1) request that DRS provides a mechanism for us to coexist through one of the options I present here.  So you can use nested bundles and we can use a "flat" bundle with "/" in the file names (and we're happy to encode these "/" so you and other DRS clients can simply ignore them and not need to create directory structures when downloading from the HCA DRS repo).  Or 2) not support DRS and go "off spec" with our implementation.  I'd really not like option 2 here since I want all DRS implementations to fully support the spec.

Let's discuss today!!

junjun-zhang:
@briandoconnor, your point regarding HCA implementation is well taken. Indeed, there are two ways (maybe more) to implement directory support. As anything else, there are pros and cons to each of them. Yes, let's discuss.

susheel:
A couple of points:
* The original Bundle/Collections discussion went into the expansion of bundle. See https://github.com/ga4gh/data-repository-service-schemas/issues/221
* Although the current focus of the discussion is on forward-slashes. The PR as it stands needs some work as whitespaces are allowed chars on Unix filesystems and need to be supported.

xbrianh:
The HCA does not support nested bundles, and allows slashes in file names within bundles. We strongly prefer option 3, "slashes allowed and interpreted by a client", as presented by @briandoconnor [above](https://github.com/ga4gh/data-repository-service-schemas/pull/276#issuecomment-500291008).

briandoconnor:
As a followup to this, today we discussed an approach that would satisfy both the HCA and ICGC/Elixir use cases.  The idea is to have the ability for a given bundle to return, in a single response, all the nested bundles and their contents listing.  With the HCA interpreting this as a bundle and all the sub directories encoded in file names and Elixir/ICGC interpreting this as sub-bundles and their content.  So, basically, a win-win for all groups and for clients that, I'm sure, will like having a consistent way to represent nested bundles/sub-folders.

@ddietterich and @mikebaumann said they would put together a PR for a potential response for this.

junjun-zhang:
@briandoconnor, sorry I wasn't able to attend today's call. I like the idea to come up with a win-win solution.

Just a thought here: how about adding an optional field explicitly for `path`? instead of overloading the `name` field which as I understand is for the `basename` of a file. If `path` is missing, the client interprets it as the current folder. This way an implementer may choose either approach to support folder structure. In fact, it would work if one chooses (although unlikely) to have both at the same time: bundle/sub-bundle and `path`.

Regarding the difficulty exploring full hierarchical folder structure when implemented via bundle/sub-bundle, I think there is a common agreement among us that there will be an index component on top of DRS to provide much more performant and sophisticated search capability.

briandoconnor:
With #278 I think HCA is a +1 for this.  @xbrianh and @mikebaumann hope that is OK with you both.

ddietterich:
+1 from Terra Data Repo

ddietterich:
+1 from Terra Data Repo

PR #278, opened 2019-06-11
Add expand query parameter and recursive bundle representation

ddietterich:
As discussed in our meeting of June 10, this PR adds two things:
1. A `expand` query flag on the `GET /object/{object_id}` endpoint to request a full, recursive expansion of all nested bundles, if the object is a bundle.
2. A `contents` field in the `ContentsObject` to represent the contents of the nested bundle.

I included the semantic that a server is not required to provide object ids or DRS URIs for nested bundles.

### DIFF
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -71,6 +71,22 @@ paths:
           in: path
           required: true
           type: string
+        - in: query
+          name: expand
+          type: boolean
+          default: false
+          description: >-
+            If false and the object_id refers to a bundle, then the ObjectContents array
+            contains only those objects directly contained in the bundle. That is, if the
+            bundle contains other bundles, those other bundles are not recursively
+            included in the result.
+            
+            If true and the object_id refers to a bundle, then the entire set of objects
+            in the bundle is expanded. That is, if the bundle contains aother bundles,
+            then those other bundles are recursively expanded and included in the result.
+            Recursion continues through the entire sub-tree of the bundle.
+
+            If the object_id refers to a blob, then the query parameter is ignored.
       tags:
         - DataRepositoryService
       x-swagger-router-controller: ga4gh.drs.server
@@ -361,10 +377,22 @@ definitions:
           A list of full DRS identifier URI paths
           that may be used to obtain the object.
           These URIs may be external to this DRS instance.
+
+          If this ContentsObject describes a nested bundle, the drs_uri array may be
+          empty. A DRS server need not provide an object id or DRS URI for a nested bundle.
         example:
           drs://example.com/ga4gh/drs/v1/objects/{object_id}
         items:
           type: string
+      contents:
+        type: array
+        description: >-
+          If this ObjectContents describes a nested bundle and the caller specified
+          "?expand=true" on the request, then this contents array must be present and
+          describe the objects within the nested bundle.
+        items:
+          $ref: '#/definitions/ContentsObject'
+        
     required:
       - name
       - id
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -76,7 +76,7 @@ paths:
           type: boolean
           default: false
           description: >-
-            If false and the object_id refers to a bundle, then the ObjectsContent array
+            If false and the object_id refers to a bundle, then the ContentsObject array
             contains only those objects directly contained in the bundle. That is, if the
             bundle contains other bundles, those other bundles are not recursively
             included in the result.
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -76,7 +76,7 @@ paths:
           type: boolean
           default: false
           description: >-
-            If false and the object_id refers to a bundle, then the ObjectContents array
+            If false and the object_id refers to a bundle, then the ObjectsContent array
             contains only those objects directly contained in the bundle. That is, if the
             bundle contains other bundles, those other bundles are not recursively
             included in the result.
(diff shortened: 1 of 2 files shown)


openapi/data_repository_service.swagger.yaml
```diff
@@ -361,10 +377,22 @@ definitions:
...
+          If this ContentsObject describes a nested bundle, the drs_uri array may be
+          empty. A DRS server need not provide an object id or DRS URI for a nested bundle.
         example:
           drs://example.com/ga4gh/drs/v1/objects/{object_id}
         items:
           type: string
+      contents:
+        type: array
+        description: >-
+          If this ObjectContents describes a nested bundle and the caller specified
```
dglazer: s/ObjectContents/ContentsObject/

openapi/data_repository_service.swagger.yaml
```diff
@@ -361,10 +377,22 @@ definitions:
           A list of full DRS identifier URI paths
           that may be used to obtain the object.
           These URIs may be external to this DRS instance.
+
+          If this ContentsObject describes a nested bundle, the drs_uri array may be
+          empty. A DRS server need not provide an object id or DRS URI for a nested bundle.
```
dglazer: I don't think this comment is needed, since I believe the `drs_uri` array is always optional.

dglazer:
Thanks Dan -- this came together even more cleanly than I expected; eager to hear if it works for the HCA use case as well as we hope.

I added a couple of minor comments on the comments

junjun-zhang:
I see a potential problem with this: what if a bundle contains millions sub-bundle/blob objects?

Here is from @briandoconnor's [comment](https://github.com/ga4gh/data-repository-service-schemas/pull/276#issuecomment-500473498):
> in HCA we have bundles that contain millions of files, all with directory structures

A GET request on such a bundle may end up getting a JSON response that is prohibitively large. Because the response is a single JSON object, it's not even possible to break it up into multiple requests/responses via pagination.

I am not sure what alternative design choices could help with this in DRS. Maybe one option is to leave this type of search functionality for another component such as an index/search service.

openapi/data_repository_service.swagger.yaml
ddietterich: Good catch, will fix.

openapi/data_repository_service.swagger.yaml
ddietterich: I will add the id field comment as you suggest. Here I was trying to capture the intent from our discussion.

I am curious what the rationale is for this to be optional in the non-nested case. It seems to say: there exists an object foo, but there is no way to access it.

ddietterich:
@junjun-zhang I think that problem already existed. If a bundle has millions of things in it, we have a very large response and no way to page it.

The alternative is between a flat list with implied directories or a tree with specified directories. Size-wise, I don't think it materially changes the problem you identify.

dglazer:
I think the id is sufficient to fetch the object, since you already know
the server. I don't remember the rationale for also allowing the uri --
maybe it's in case there are copies of the object on other servers? Susheel
probably remembers, or I can dig through old comments

ddietterich:
@dglazer ah, I get it. The URIs are really a short-cut to save a caller from issuing GET /object for each id. I'll remove my description text. Thanks!

junjun-zhang:
@ddietterich you are right, the problem already exists. What came across my mind was the difference among `ls`, `ls -R`, `ls /` and `ls -R /`. Indeed if there are millions of things **directly** under a folder/bundle (although I don't know what's the use case for that, in ICGC we will definitely avoid it), the response will still be large even for `ls`. Good thing the `expand` param defaults to `false`.

I guess an implementer could choose not to support `expand`, is that possible? In our case, `ls /` is fine but definitely not for `ls -R /`

ddietterich:
My understanding is that HCA has bundles with millions of objects in a hierarchy and need a way to return the entire hierarchy. The choice is between two output formats of `ls -R`

One option is putting the millions of things in one list with the hierarchy implied by names like `/top/second/third/file`.

The other choice is to represent the things as hierarchy of lists `top: [ second: [ third: [ file ] ] ]` as this PR proposes.

Either way you have millions of things in the output.

geoffjentry:
@junjun-zhang I'd like to caution against relying too much on a potential indexing/search service for things like this. Imagine the use case of a workflow system which is passed DRS URLs as file inputs. Presumably the underlying system will need to retrieve all the relevant files associated with those URLs

junjun-zhang:
@geoffjentry thanks for pointing it out, I do have this primary use case in mind. This is why I agree with @susheel regarding [accessing input file](https://github.com/ga4gh/data-repository-service-schemas/pull/271#issuecomment-500206132) in the form of `/<bundle_id>/<object_id>` (or even `/<bundle_id>/<file_name>` as @dglazer [pointed out later](https://github.com/ga4gh/data-repository-service-schemas/pull/271#issuecomment-500386047)), so that workflow system will be able to figure out secondary files without relying on anything else.

Just to be clear I am not against this PR, rather just want to bring up the potential problem of large response size, and I am happy to hear how experts in this group think about it.

briandoconnor:
Sorry for the duplicates, the wifi in my hotel is causing problems and resulted in multiple submissions

briandoconnor:
@mikebaumann and @xbrianh I think this will allow us to represent bundled with "/" as nested arrays in the response (which is better than not having support for HCA's approach in DRS).  There may be ways we have to improve this in the future but I think HCA should give a +1 for this so we can have some way to represent our bundles with "/" in the file names in 1.0 without going off spec.  Do you agree?

I'll vote +1 for HCA unless I hear a negative for you

dglazer:
Discussed on today's Cloud WS call -- although this is moving fast, we feel comfortable merging, because we know it addresses both HCA's and Elixir's needs for bundles in a single clean way.

mikebaumann:
Post-merge comment:
I think this is a clean unification of the output format for the two disparate underlying models of 1) single bundle with nested files (i.e. HCA), and 2) nested bundles.
For the first case (e.g. HCA), my understanding is the expand option has no effect, all files are in a single bundle and will therefore always be listed. I think HCA can support this operation once the pagination/scalability issues of the current definition have been addressed.
It is probably worth opening an issue regarding that now.

PR #281, opened 2019-06-27
Add self property drs object

patmagee:
## Problem
I am in the process of creating a number of  POC services which consume a `DRS` URI (using the develop schema) as input in order to demonstrate the capabilities of `DRS`. Throughout this process I have come across a few situations where I am left with a `DRS` object but have lost the initial context from which it was retrieved from. The technical reason why I have encountered this issue is due to the separation of concerns. The component which retrieves the `DRS` object is separate then the one which actually does some down stream processing of it. 

In most situations, not knowning where the object came from is fine. However If ever I need to hit the `objects/{id}/access/{access_id}`, I am unable to resolve the appropriate path to retrieve the `access_url`. I do believe there will be many other scenarios for consumers of `DRS` where an `object` becomes detached from the context which it was retrieved, rendering any additional API calls impossible.

## Proposal
Fundamentally, I believe a `DRS` object should be self describing, including where the object came from.   I propose the inclusion of a single `self` property within the `Object`. The property should be a `DRS` `URI` which refers to the exact location which produced said object.

Ideally, this property would be required, however I understand there may be some scenarios where there is no `self` and the `DRS` object is just a wrapper to convey information on an object.

### DIFF
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -217,6 +217,12 @@ definitions:
         description: |-
           For blobs, the blob size in bytes.
           For bundles, the cumulative size, in bytes, of items in the `contents` field.
+      self:
+        type: string
+        description: |-
+          A DRS URI referencing where this `Object` was retrieved from.
+        example:
+          drs://example.com/ga4gh/drs/v1/objects/{id}
       created:
         type: string
         format: date-time


sarpera:
Perhaps consider "self" to be a DRS ID instead of DRS URI? So that there wouldn't be a need to update the value for URI in case the supported version or domain name changes.

ddietterich:
@sarpera can you explain further?

I don't think having just the DRS ID it solves the stated problem of being able to re-retrieve the object. You need the server DNS to know who to ask to resolve

I am skeptical of "the domain name changes" use case. DNS names can be logical, so the server can move without a name change.

If a DRS server stops supporting a URI version, then has it not orphaned all persisted DRS URIs of that version?

sarpera:
@ddietterich @patmagee I'm sorry, I misread the changed file in the PR thinking that the suggested `self` property was meant for `objects/{id}/access/{access_id}` call. 

Regardless of this PR, 
> If a DRS server stops supporting a URI version, then has it not orphaned all persisted DRS URIs of that version?

Yes, this is the expected behaviour. This PR doesn't conflict with that, my bad misinterpreting the  code changes.

patmagee:
@sarpera No problem at all. You are indeed correct, resolving the `access_id` is my primary motivator for this.

dglazer:
I apologize for missing this discussion earlier. I just created #284 to address the same issue (which was also raised by our Product Review Committee), but with a slightly different syntax, and didn't see this PR until afterwards.

The main difference is that I propose using a "drs://" URI instead of a directly resolvable HTTP URL. That format is our recommended way to pass around text references to objects, and I believe it's slightly more likely to remain valid over time.

patmagee:
@dglazer thanks for bringing this up at the committee. The changes highlighted above specify that the `self` link MUST be a `drs` uri already, so it would seem as though our PR's are mostly duplicates of one another! it looks like your PR has more traction, so please feel free to close this!

dglazer:
Sounds good @patmagee  -- closing.

PR #282, opened 2019-08-03
Tighten up checksum types 

dglazer:
Response to [DRS v1 PRC #5 - Type Checksum](https://docs.google.com/document/d/1gCTFFomdUhP7dPJD6H-eqQ77OmLszvKGDqJXaBkU9SU/edit#heading=h.h385fnqzbzmk) feedback

### DIFF
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -191,7 +191,7 @@ definitions:
         description: |-
           The digest method used to create the checksum.
 
-          The value (e.g. `sha-256`) SHOULD be listed in the https://www.iana.org/assignments/named-information/named-information.xhtml#hash-alg[IANA Named Information Hash Algorithm Registry].
+          The value (e.g. `sha-256`) SHOULD be listed as `Hash Name String` in the https://www.iana.org/assignments/named-information/named-information.xhtml#hash-alg[IANA Named Information Hash Algorithm Registry].
           Other values (e.g. `md5`, `etag`) MAY be used, as long as implementors are aware of the issues discussed in https://tools.ietf.org/html/rfc6920#section-9.4[RFC6920].
   Object:
     type: object
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -188,14 +188,17 @@ definitions:
         description: 'The hex-string encoded checksum for the data'
       type:
         type: string
-        description: |-
+        description: >-
           The digest method used to create the checksum.
+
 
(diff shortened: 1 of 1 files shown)


dglazer:
Thanks for the extra specificity @susheel

susheel:
+1 ELIXIR

ddietterich:
+1 Terra

Avantol13:
+1 DCF/UChicago/CTDS/Gen3

mattions:
+1 from Seven Bridges (CGC, Cavatica)

susheel:
Agree. In addition, I think it might be worth standardising this into a GA4GH specific hash algorithm registry by extending the [IANA Hash Algorithm Registry](https://www.iana.org/assignments/named-information/named-information.xhtml#hash-alg)

See: https://github.com/susheel/ga4gh-hash-alg-registry/blob/master/hash-alg.csv

This could be ideally be forked and maintained by GA4GH Discovery.

briandoconnor:
Ya @susheel that's a really good point.

@mfiume and @mbaudis could you guys take this on?  I think it would be extremely helpful to have all the checksum algorithm names consistent across GA4GH APIs.

dglazer:
Potential language to discuss on today's call:
```
description: |-
          The digest method used to create the checksum.
          The value (e.g. `sha-256`) SHOULD be listed as `Hash Name String` in the https://www.iana.org/assignments/named-information/named-information.xhtml#hash-alg[IANA Named Information Hash Algorithm Registry].
          Other values MAY be used, as long as implementors are aware of the issues discussed in https://tools.ietf.org/html/rfc6920#section-9.4[RFC6920].
          GA4GH may provide more explicit guidance for use of non-IANA-registered algorithms in the future.
          Until then, if implementors do choose such an algorithm (e.g. because it's implemented by their storage provider), they SHOULD use an existing standard type such as `md5` or `etag`.
```
I have mixed feelings about this approach -- on the one hand it reflects reality; on the other hand it encourages exactly the sort of behavior the PRC was trying to discourage. I think SHOULD/MAY covers it, but am open to leaning harder either way.

dglazer:
Notes from today's call:
- this language strikes a decent balance between too loose and too tight
- add `crc32c`, `trunc512`, `sha1`

mfiume:
> Ya @susheel that's a really good point.
> 
> @mfiume and @mbaudis could you guys take this on? I think it would be extremely helpful to have all the checksum algorithm names consistent across GA4GH APIs.

I'm okay with this. I have forked it into Discovery's organization, [here](https://github.com/ga4gh-discovery/ga4gh-hash-alg-registry), and made @susheel the admin. Susheel, can you please add a README?

Issue #11, opened 2017-10-17
Bundle Alias


david4096:
From @bwalsh

> Do we need an alias of the bundle? Wouldn't one just create a second bundle?

david4096:
Aliasing bundles let's institutions have an easy to pass around value that can be used to reliably get at the data. So I vote keep it.

david4096:
Closing as this is present in the current release https://github.com/ga4gh/data-object-service-schemas/releases/tag/0.1.0

PR #283, opened 2019-08-03
remove /service-info endpoint (for now)

dglazer:
Addresses [DRS v1 PRC #11 /service-info items](https://docs.google.com/document/d/1gCTFFomdUhP7dPJD6H-eqQ77OmLszvKGDqJXaBkU9SU/edit#heading=h.y5dqqkkd2lp4)

Per PRC discussion, the GA4GH-wide /service-info spec isn't yet ready for use.

### DIFF
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -21,20 +21,6 @@ security:
   - {}
   - authToken: []
 paths:
-  '/service-info':
-    get:
-      summary: Get information about this implementation.
-      description: >-
-        May return service version and other information.
-      operationId: GetServiceInfo
-      responses:
-        '200':
-          description: Service information returned successfully
-          schema:
-            $ref: '#/definitions/ServiceInfo'
-      tags:
-        - DataRepositoryService
-      x-swagger-router-controller: ga4gh.drs.server
   '/objects/{object_id}':
     get:
       summary: Get info about an `Object`.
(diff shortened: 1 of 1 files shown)


susheel:
Addresses [DRS v1 PRC #11 /service-info items](https://docs.google.com/document/d/1gCTFFomdUhP7dPJD6H-eqQ77OmLszvKGDqJXaBkU9SU/edit#heading=h.y5dqqkkd2lp4)

susheel:
I don't think this endpoint needs deletion for DRS v1, as points raised need to be addressed by Discovery and future inclusion post v1

dglazer:
I'm personally fine either way, so unless the PRC has a strong opinion, I defer to the group. The bigger argument *for* deleting it now is that, once there's a GA4GH-blessed cross-API /service-info definition, we can add it to DRS without a backwards-incompatible breaking change.

susheel:
I'd keep the `/service-info` endpoint to at least provide users links to the documentation and contact points of the service, including being consistent with all other GA4GH APIs. Not a strongly held opinion and happy to defer to the community.

Once we get a GA4GH blessed `/service-info` spec, I suspect this will be a breaking change for all GA4GH APIs.

susheel:
+0 ELIXIR

ddietterich:
+1 Terra
We'll support it as an extension and then totally replace it when the real thing comes along. No backward compatibility.

Avantol13:
+1 from DCF/UChicago/CTDS/Gen3/<insert other identifiers here>

mattions:
+0 from Seven Bridges (CGC, Cavatica)
(Agreed that vocabulary should be shared among other Cloud WS APIs. Currently it might be useful to link to doc pages)

PR #284, opened 2019-08-03
Add Object.self_uri

dglazer:
Address [DRS v1 PRC #2 - Add a self link to DRS objects](https://docs.google.com/document/d/1gCTFFomdUhP7dPJD6H-eqQ77OmLszvKGDqJXaBkU9SU/edit#heading=h.q13g7vd8vdmf) feedback by adding a self_uri field to DRS objects

Related PR: #281

### DIFF
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -216,6 +216,8 @@ definitions:
         description: |-
           A drs:// URI, as defined in the DRS documentation, that tells clients how to access this object.
           The intent of this field is to make DRS objects self-contained, and therefore easier for clients to store and pass around.
+        example:
+          drs://drs.example.org/314159
       size:
         type: integer
         format: int64
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -214,7 +214,7 @@ definitions:
       self_uri:
         type: string
         description: |-
-          A drs: URI, as defined in the DRS documentation, that tells clients how to access this object.
+          A drs:// URI, as defined in the DRS documentation, that tells clients how to access this object.
           The intent of this field is to make DRS objects self-contained, and therefore easier for clients to store and pass around.
       size:
         type: integer
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -200,7 +200,7 @@ definitions:
           sha512
   Object:
     type: object
-    required: ['id', 'size', 'created', 'checksums']
+    required: ['id', 'self_uri', 'size', 'created', 'checksums']
     properties:
       id:
         type: string
@@ -211,6 +211,11 @@ definitions:
         description: |-
           A string that can be used to name an `Object`.
           This string is made up of uppercase and lowercase letters, decimal digits, hypen, period, and underscore [A-Za-z0-9.-_]. See http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_282[portable filenames].
+      self_uri:
+        type: string
+        description: |-
+          A drs: URI, as defined in the DRS documentation, that tells clients how to access this object.
+          The intent of this field is to make DRS objects self-contained, and therefore easier for clients to store and pass around.
       size:
         type: integer
         format: int64


susheel:
I'm guessing the URI part of the drs://URI points to the full URI path, i.e.
drs://server.com/65a3e501-7448-4887-aada-b52de5c8db31

susheel:
Address [DRS v1 PRC #2 - Add a self link to DRS objects](https://docs.google.com/document/d/1gCTFFomdUhP7dPJD6H-eqQ77OmLszvKGDqJXaBkU9SU/edit#heading=h.q13g7vd8vdmf) feedback

dglazer:
Thanks for the slashes @susheel

susheel:
+1 ELIXIR

ddietterich:
+1 Terra

Avantol13:
+1 from DCF/UChicago/CTDS/Gen3/<insert other identifiers here>

pr80ik:
+1

mattions:
-0 from Seven Bridges (CGC, Cavatica)

self URI, ie DRS URL, contains server version info which might violate the principle that a URI should refer to a unique resource. This PR would be a +1 from us if DRS URLs in general wouldn't include versions, but instead rely on request headers, but if we feel we have to go this one, will tag along.

susheel:
I think the spirit of this PR is to provide a field that refers to the URI that generated the current DRS response. It will be volatile but is intended to help downstream services to obtain the DRS server URI to help resolve the object's access_urls.

The `self_uri` should follow the convention noted the [DRS URIs documentation](https://github.com/ga4gh/data-repository-service-schemas/blob/release/drs-1.0.0/docs/asciidoc/front_matter.adoc#L21) i.e. `drs://drs.example.org/314159`

It would be up to the DRS client to turn this `self_uri` into the appropriate HTTP request as documented [here](https://github.com/ga4gh/data-repository-service-schemas/blob/release/drs-1.0.0/docs/asciidoc/front_matter.adoc#L21).

We could add this example to help clarify this PR.

sarpera:
@susheel thanks for the clarity, I think everyone is onboard with the intention of this PR as long as DRS URIs don't contain server version info. 

From the [docs](https://github.com/ga4gh/data-repository-service-schemas/blob/release/drs-1.0.0/docs/asciidoc/front_matter.adoc#drs-uris):
> For example, if a WES server was asked to process `drs://drs.example.org/314159`, it would know that it could issue a GET request to `https://drs.example.org/ga4gh/drs/v1/objects/314159` to learn how to fetch that object.

We might want to address the issues with having the server version in the URLs in another PR and perhaps suggest using common practices for [API versioning](https://restfulapi.net/versioning/).

dglazer:
@susheel -- thanks for the `example:`

@sarpera -- I *think* you're saying that you're happy with this PR's use of the version-free DRS URI format (e.g. `drs://drs.example.org/314159`), but would prefer that DRS method URLs themselves (e.g. `https://drs.example.org/ga4gh/drs/v1/objects/314159`) were also version-free. If so, I agree we should start that conversation in a new Issue for post-v1, and ideally come up with an approach that can be standardized across all GA4GH APIs.

@Avantol13 -- Similarly, I *think* you're saying that you're okay with this PR's use of the standard DRS URI format, but would prefer that the format itself were different (e.g. `drs:314159`). If so, I think we should have that conversation in a separate Issue for post-v1, where we can work through the tradeoffs you raise, and specific proposals for addressing. (I agree that a global namespace has advantages, and I agree it would take a lot of effort to get right.)

briandoconnor:
OK, so it looks like folks are generally OK with this provided the URI is a URL of the form described in the spec e.g. `drs://drs.example.org/314159` vs the actual URL you use to GET the object JSON:

> For example, if a WES server was asked to process drs://drs.example.org/314159, it would know > that it could issue a GET request to http://drs.example.org/ga4gh/drs/v1/objects/314159 to learn > how to fetch that object.

Just to address the totally valid concerns from @Avantol13 , this is not the final say in how DRS uses URI.  It doesn't prevent us from supporting, in a future iteration of DRS, the ability to have drs-guid://<data_GUID> for example as an additional way of specifying DRS objects using a GUID.  And we could work out a scheme for how we reference GUIDs of different types (e.g. Data GUIDs, ARKS, etc).

@Avantol13 the fact that we didn't tackle this in 1.0 came down to time.  Like you said, going with the GUID approach has many benefits. But I think we need more time (and your help along with others on this thread) for thinking through how resolution of DRS GUIDs would work in a distributed way.  It would be wonderful to see you raise a pull request to kick off a proposal around this for 1.x of DRS.

On a related note, there was some concern in this thread about how to turn a URI like `drs://drs.example.org/314159` into a URL that a client can GET on.  Right now we have one version of DRS so it's easy to turn this into http://drs.example.org/ga4gh/drs/v1/objects/314159.  However, in future releases of the specification this isn't going to be obvious.  I think we need to work with the Discovery group such that, if I have `drs://drs.example.org/314159` I can make a request to, say, `https://drs.example.org/service-info` and have that give me clear information on the DRS endpoints and versions available on this service.

patmagee:
+1 from DNAstack

PR #285, opened 2019-08-03
Clarify uniqueness of access_id

dglazer:
Addresses [DRS v1 PRC #7 - Make the uniqueness of the access_id optional](https://docs.google.com/document/d/1gCTFFomdUhP7dPJD6H-eqQ77OmLszvKGDqJXaBkU9SU/edit#heading=h.hjuct0b5v60)

### DIFF
--- a/openapi/data_repository_service.swagger.yaml
+++ b/openapi/data_repository_service.swagger.yaml
@@ -347,7 +347,7 @@ definitions:
         type: string
         description: >-
           An arbitrary string to be passed to the `/access` method to get an `AccessURL`.
-          This string must be unique per object.
+          This string must be unique within the scope of a single object.
           Note that at least one of `access_url` and `access_id` must be provided.
       region:
         type: string


susheel:
Addresses [DRS v1 PRC #7 - Make the uniqueness of the access_id optional](https://docs.google.com/document/d/1gCTFFomdUhP7dPJD6H-eqQ77OmLszvKGDqJXaBkU9SU/edit#heading=h.hjuct0b5v60)

susheel:
I feel the [DRS v1 PRC #7](https://docs.google.com/document/d/1gCTFFomdUhP7dPJD6H-eqQ77OmLszvKGDqJXaBkU9SU/edit#heading=h.hjuct0b5v60) comment was to loosen the need to enforce `access_id` to be unique per object, not tighten.

The loosening was to let providers decide how they would want to provide `access_ids` - unique per object or globally and should not be an enforced property of the standard.

dglazer:
The intent of this edit is to clarify that the requirements **are** loose -- if that's not clear, I'm wide open to different words. Providers are of course welcome to provide globally unique `access_id` values, since those also automatically meet the requirement of being unique per object.

ddietterich:
I'll support this as is, but the phrase "unique per object" does not clarify much for me.
Must access ids be unique in time? Or could they be unique until the access request is completed or reused if the object is deleted?

Avantol13:
+1 from DCF/UChicago/CTDS/Gen3, I think this wording makes sense with the assumption that we mean you can't have multiple of the same access ids as access methods for DRS ID A, but you _can_ have the same access id as access methods _across_ DRS IDs (e.g. one in A and one in B)

pr80ik:
+1

mattions:
+0

briandoconnor:
This looks reasonable to me and I see folks are either +0 or +1.

mfiume:
+1

Issue #294, opened 2019-08-21
Updates after GA4GH AAI standard is confirmed


susheel:
Need to update our documentation after the [GA4GH AAI standard](https://github.com/ga4gh/data-security/blob/master/AAI/AAIConnectProfile.md) is confirmed.

- [ ] Update front_matter.adoc with appropriate changes
> GA4GH recommends the use of the OAuth 2.0 framework [[RFC 6749](https://tools.ietf.org/html/rfc6749)] for authentication and authorization. It is also recommended that implementations of this standard implement and follow the [GA4GH Authentication and Authorization Infrastructure (AAI) standard](https://github.com/ga4gh/data-security/blob/master/AAI/AAIConnectProfile.md).

PR #295, opened 2019-08-21
Updates for after GA4GH AAI is confirmed

susheel:
- [x] Update front_matter.adoc with appropriate changes
> GA4GH recommends the use of the OAuth 2.0 framework [[RFC 6749](https://tools.ietf.org/html/rfc6749)] for authentication and authorization. It is also recommended that implementations of this standard implement and follow the [GA4GH Authentication and Authorization Infrastructure (AAI) standard](https://github.com/ga4gh/data-security/blob/master/AAI/AAIConnectProfile.md).

resolves: #294

### DIFF
--- a/docs/asciidoc/front_matter.adoc
+++ b/docs/asciidoc/front_matter.adoc
@@ -41,7 +41,7 @@ The DRS API specification is written in OpenAPI and embodies a RESTful service p
 
 === Making DRS Requests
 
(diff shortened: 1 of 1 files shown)


dglazer:
I think we still want to add references to AAI, now that it's confirmed. (Although we don't want to reference /service-info yet, so we should hold off merging until that's resolved.)

Issue #215, opened 2019-01-28
"UUID" and corresponding query endpoint


uniqueg:
### Synopsis
A strong case can be made for the usefulness of providing means to uniquely identify data objects even in `DRS v1.0`. Assigning a hash sum-based "universally unique identifer" (UUID; quoted because it will very likely not conform to [RFC 4133](https://tools.ietf.org/html/rfc4122)) to each data object (probably in addition to a less unwieldy "locally unique" identifier) may provide such a means.

### Use cases
- **Identify data**: Sometimes it happens that one is confronted with a piece of data that has no header/description/metadata attached. If the data had previously been deposited to a DRS instance that the user has access to, a call to the discovery service could quickly identify the data via the UUID query endpoints of known DRS instances and provide metadata.
- **Ensure/avoid redundancy**: For data safety and even political reasons, it is desirable to ensure that important data exists in multiple copies. On the other hand, in order to save resources or for reasons of data privacy, it is desirable to minimize the level of redundancy according to the policies of a specific data provider/vendor. UUIDs would help to enforce such data redundancy policies by implementing a mechanism by which copies of data objects can be found and annotated automatically, either within individual DRS instances (via queries to the discovery service) or within the discovery service itself (centralized database).
- **Efficient use of caches**: Data consumers may wish to analyze a given data object (e.g. when discovering a DRS ID associated with a publication) that is already available in a local cache. In most cases, one would wish to analyze the local copy of the data object as it would likely speed up the analysis and minimize costs. UUIDs could be easily leveraged by WES implementations to check whether the data a user wants to analyze is available in a local cache.
- **Optimize distributed computing**: Allowing to quickly and unambiguously identify all accessible copies of a given data objects (given it's DRS identifier) would enable WES/TES to make more informed decisions on distributing (parts of) analysis workflows according to considerations such as data location, access rights/use restrictions, load balance and costs.
- **Federation of access rights**: If a data consumer requests a data object that they have access rights to on at least one mirror, data providers/vendors may choose to propagate these permissions for the same data object under their own DRS instance, even though the user does not have direct access rights to that data object on that platform. UUIDs could provide a mechanism to ensure that data consumers have universal access to all copies of a data object they have access to.

### Implementation
While the assignment of random, data-independent UUIDs would be computationally cheap, this mechanism is rather unfeasible, as UUID assignment would not be consistent for multiple copies of the same data object without resorting to centralized or distributed (e.g., blockchain) data registries.

[...]

PR #312, opened 2020-03-06
Drs 306 persistent identifiers and prefixes

MichaelLukowski:
Adding functionality for prefixes to allow for persistent DRS URLs.


dglazer:
@MichaelLukowski -- have you looked into why the CI builds are failing? It would be easier to review if we could look at the generated documentation.
@jaeddy -- is this possibly related to your work in progress on improving the build infra?

docs/asciidoc/front_matter.adoc
```diff
@@ -63,3 +72,33 @@ The DRS API allows implementers to support a variety of different content access
...
+A DRS Resolution flow would look like this:
+
+Propose we have a DRS URL to resolve in the form of <drs://example.com/ga4gh/drs/v1/object/dg.4503/182>
+
+Send a GET request to the DRS URI
+	e.x. (GET http://example.com/ga4gh/drs/v1/object/dg.4503/182)
+
+If this returns a 404 error we can then resolve this guid to find the appropriate server URL
+	If the prefix starts with `dg.` we then use dataguids.org as the resolver and send a new GET request
+		GET https://dataguids.org/ga4gh/drs/v1/dataobjects/dg.4503/dg.4503/182
```
dglazer: I'm uncomfortable with a GA4GH spec including a hard-wired mapping to a project-specific resolver (e.g. "dg --> dataguids.org"). It will look like playing favorites, and it will open us up to never-ending requests to add new project-specific mappings to the spec, which every DRS client would then have to implement.

Can we remove the hard-wired reference, and instead take advantage of existing widely-available general-purpose services?

docs/asciidoc/front_matter.adoc
```diff
@@ -63,3 +72,33 @@ The DRS API allows implementers to support a variety of different content access
...
 
 DRS implementers should ensure their solutions restrict access to targets as much as possible, detect attempts to exploit through log monitoring, and they are prepared to take action if an exploit in their DRS implementation is detected.
+
+=== DRS Resolution
+
+When making DRS requests if a 404 is encountered, the DRS implementation CAN use the DRS prefix to resolve the identifiers. DRS implementors can use a resolution service like http://n2t.net/[n2t], http://identifiers.org/[identifiers.org], or https://dataguids.org/[dataguids] (for `dg.`) to resolve prefixes. This request will give a HTTP 302 response with the redirect to the new URL associated with the DRS implementation. This new URL can then be cached to be used for subsequent DRS requests.
+
+A DRS Resolution flow would look like this:
+
+Propose we have a DRS URL to resolve in the form of <drs://example.com/ga4gh/drs/v1/object/dg.4503/182>
```
dglazer: minor: change "Propose" to "Suppose"

docs/asciidoc/front_matter.adoc
```diff
@@ -63,3 +72,33 @@ The DRS API allows implementers to support a variety of different content access
...
+
+Send a GET request to the DRS URI
+	e.x. (GET http://example.com/ga4gh/drs/v1/object/dg.4503/182)
+
+If this returns a 404 error we can then resolve this guid to find the appropriate server URL
```
dglazer: I don't understand how this works. If the server at example.com is working well enough to return a 404, why doesn't it just return the DRS object? If we're trying to solve cases where example.com has gone out of business, we won't be getting a 404. (Plus, how would a client be able to tell the difference between "this object doesn't exist" and "this server has moved"?)

openapi/data_repository_service.swagger.yaml
```diff
@@ -21,7 +21,7 @@ security:
   - {}
   - authToken: []
 paths:
-  '/objects/{object_id}':
+  '/objects/{scheme}{identifier}/{object_id}':
```
dglazer: I think this line means that this PR is a 100% breaking change, since all DRS requests going forward MUST include an extra path component before the `object_id` -- is that right?

docs/asciidoc/front_matter.adoc
dglazer: Why is `dg.4503` repeated twice in this path?

docs/asciidoc/front_matter.adoc
```diff
@@ -63,3 +72,33 @@ The DRS API allows implementers to support a variety of different content access
...
+	If the prefix starts with `dg.` we then use dataguids.org as the resolver and send a new GET request
+		GET https://dataguids.org/ga4gh/drs/v1/dataobjects/dg.4503/dg.4503/182
+
+	If the prefix starts with `ark:` we then can use n2t.net as the resolver and send a new GET request
+		GET http://n2t.net/ark:67531
+
+	If these requests fail you may try identifiers.org as a resolver. As it is a central resolver that handles many prefixes.
+		GET http://identifiers.org/dg.4503 or GET http://identifiers.org/ark:67531
```
dglazer: These examples show that calls to the resolver only pass the prefix. The `dataguids.org` example above, however, passes in the full URL. Which is correct.?

docs/asciidoc/front_matter.adoc
```diff
@@ -63,3 +72,33 @@ The DRS API allows implementers to support a variety of different content access
...
+
+	If all of these fail then this DRS URI cannot be resolved
+
+If these requests return a 302 with the correct URL to send a new GET request to. A DRS implementation CAN cache this new URL and use it until it doesn't work any more
```
dglazer: Is 302 the only valid response for a resolver to send, and everything else should be considered a failure? Or are there other valid responses?

dglazer: What scope of caching is the DRS client allowed to use? Can the client assume that all ids with a given prefix should be resolved using the new URL? Or does it need to separately cache the URL for each object id?

docs/asciidoc/front_matter.adoc
```diff
@@ -63,3 +72,33 @@ The DRS API allows implementers to support a variety of different content access
...
+	GET http://<NEW URL>/dg.4503/182
+
+There is no need to pass any authorization headers to the resolution service
```
dglazer: Three security questions: 

1. I'd like us to be more prescriptive here -- clients should either be required or forbidden to pass auth headers to resolvers. I think the answer is forbidden, since it's a security risk to pass bearer tokens to unknown servers.

1. How are resolvers secured? The examples show `http:` URLs, which suggests they're not, which feels like it could open us up to attackers hijacking DRS requests by taking over a resolver.

1. What are we doing to protect against eavesdropping on traffic patterns, where anyone who can watch the traffic to a resolver can find out which clients are making which DRS requests on which objects?

dglazer:
@MichaelLukowski , I just left a bunch of comments in the code. I think many of them also apply to #313 -- would you like me to add them there also, or should we work through most of them here, and then only discuss PR-specific questions in the other one?

docs/asciidoc/front_matter.adoc
```diff
@@ -63,3 +72,33 @@ The DRS API allows implementers to support a variety of different content access
...
+=== DRS Resolution
+
+When making DRS requests if a 404 is encountered, the DRS implementation CAN use the DRS prefix to resolve the identifiers. DRS implementors can use a resolution service like http://n2t.net/[n2t], http://identifiers.org/[identifiers.org], or https://dataguids.org/[dataguids] (for `dg.`) to resolve prefixes. This request will give a HTTP 302 response with the redirect to the new URL associated with the DRS implementation. This new URL can then be cached to be used for subsequent DRS requests.
+
+A DRS Resolution flow would look like this:
+
+Propose we have a DRS URL to resolve in the form of <drs://example.com/ga4gh/drs/v1/object/dg.4503/182>
+
+Send a GET request to the DRS URI
+	e.x. (GET http://example.com/ga4gh/drs/v1/object/dg.4503/182)
```
dglazer: minor: change e.x. to e.g.

docs/asciidoc/front_matter.adoc
dglazer: minor: the first sentence isn't quite grammatically correct. I believe the intent is:

If these requests return a 302, then the `Location` header in the response contains the URL of the DRS server to which requests should be sent.

ddietterich:
I would like to understand more clearly the requirements and use cases for these changes. I think what is proposed might be too narrowly drawn. Perhaps there are more general cases and a more general solution.

From the perspective of Terra Data Repository, we would not expect to use an external archival service. We own the responsibility for geo-placement, security, and access control of the data we have been entrusted. I doubt we would pass that responsibility off to a separate entity. We plan to take responsibility for our own archival and ongoing support of the DNS names and servers that will translate them.

One way to view an archival service is as maintaining a map from DRS URIs to archived copies of the data. Suppose I have DRS URI `drs://drs.repo.bio/a8cc7c00-059c-4896-8e34-508f3d05a`. I try to resolve at drs.repo.bio and find the service no longer exists. So I send the whole URI to an endpoint at the archival service (not a DRS V1 endpoint). It maps the URI to its copy of the object and provides the object. It requires no change to existing DRS endpoints.

Another way to view an archival service is as a DRS data repository. That is the view I prefer. The archive has a DNS name (dataguids.org) and an identifier (4503). So we can compose a DRS URI for the archival object: `drs://dataguids.org/4503`. Is there any functional property that makes it different than any other DRS data repository? Or is it just promises of ongoing resources to support the data storage, DNS name, operations, etc.

Assume we treat an archival service as a DRS data repository. We don't have to invent new encodings or grant archival services special names to describe objects within them. 

I can we can restate the problem this PR tries to solve as finding a way to provide an ordered search list of (server/id) within one URI. Rather than overloading the existing DRS URI, I suggest we make a separate URI format. Something like: `drsx:server/id:server/id:server/id`. A DRS implementation could optionally support a resolving endpoint that would walk the list until it found an object. That would cover the archival use case and be extensible for other use cases.

uniqueg:
@MichaelLukowski: I'm wondering in how far (if at all) this PR and #313 may address the use cases described in #215? Could the resolvers be used to locate different copies/geolocations of the same data?

openapi/data_repository_service.swagger.yaml
briandoconnor: So if this is optional, you just want to add another endpoint rather than adding in {schema}{identifier}, right?  I don't think OpenAPI has a way of adding in an optional part of the URL path, correct?

docs/asciidoc/front_matter.adoc
ianfore: Can we review identifiers.org and n2t.net as possible resolvers?
These have increasing support and already has 690 prefixes registered.

At the same time I'm not sure why a drs: prefix makes sense vs using the prefixes that repositories have registered, or will do so.

[Relevant paper](https://www.nature.com/articles/sdata201829)
The journal also announced its adoption of the system in an [accompanying editorial](https://www.nature.com/articles/sdata201895)

openapi/data_repository_service.swagger.yaml
dglazer: Actually -- since the DRS server itself never has to process the prefix, why would we even include it in the DRS API? Why can't we leave DRS alone, and just add new rules for how DRS clients resolve DRS URI's that include prefixes?

abaumann:
Thanks for making the PR with more details, and it was really helpful to read through comments to better think through the implications of this change.  I've been aware of the why for a long time, but wondering more about the how.  I also agree with @ddietterich after reading through, it does actually seem very solvable by an archival service.  I also wonder if this could be used before the fact (before a previous hostname-based DRS URI is no longer available).  

For instance, in Terra, I'd like to maintain provenance so that when someone did an analysis today, you'd still be able to recreate it without error 2 years from now.  In order to do this, we could consider an "archival service" which might instead be thought of more like a "cache service" or "mapping service" that can take in that DRS GUID at a standard hostname (e.g. in your example dataguids.org) and could make a request to the appropriate current hostname, which is basically just doing DNS, but doesn't change the spec.  If for example a consumer of DRS was worried about the features/scalability/etc of dataguids.org, they could stand up their own mapping service to call and do whatever they want internally, but the URIs you use are still DRS compliant and just contain the hostname of this mapping service.  

I was concerned with who would actually do the resolution with this spec change (i.e. what actual service(s) like dataguids.org), who maintains that, how we know which one to use, etc.  Thinking about it like how Dan specified or how I was proposing might decouple that concern since each consumer could then choose to host their own for their own purposes, and that layer could also handle caching as needed.  Our current DRS handling implementation (which is actually DOS) in Terra actually does something like this already to handle these prefixes and says what hostname to call based upon a given prefix.

dglazer:
Building on @ddietterich and @abaumann 's comments -- my current thinking is:
- we don't need to change how DRS servers themselves work; the need is to define a way to find the correct DRS server if the initial one isn't available
- therefore, the two things to nail down are a) what info is needed in DRS URI's to help DRS clients find DRS servers  and  b) exactly how clients should use that info to find the right server

GIven that, here's a sketch of a very simple answer that I believe meets all the requirements:

1. zero changes to DRS API or to DRS URI spec (e.g. `drs://obsolete.org/314159`)
2. update the URI section of the docs to tell DRS clients that if the DRS server in the URI doesn't exist, they should
   1. call identifiers.org with `drs:servername`, which will 302 them to a new cacheable servername (e.g. `drs:obsolete.org` will redirect to `https://newhotness.org`.)
   2. resolve the DRS URI normally (e.g. as if it were `drs://newhotness.org/314159`) 

If that works, I'm happy to put it into a PR. If not, lmk the concerns -- we can certainly make it more complicated if needed.

ianfore:
I'm still not clear why a drs: prefix is needed vs using the existing prefixes (namespaces) registered by repositories.

The question of which resolver to use is a good one. Do the mechanisms for how you chose your DNS resolver give us a model for that? i.e. it's a local configuration issue, with the option to indicate alternatives, order of preference etc. A layer like that over DRS might helpful.

briandoconnor:
> @MichaelLukowski -- have you looked into why the CI builds are failing? It would be easier to review if we could look at the generated documentation.
> @jaeddy -- is this possibly related to your work in progress on improving the build infra?

This is not being built by travis because the feature branch doesn't following our naming convention.  See [CONTRIBUTING.md](CONTRIBUTING.md)

docs/asciidoc/front_matter.adoc
mbdebian: Hi @dglazer,
This is Manuel from [identifiers.org](https://identifiers.org).
When using the resolution services, in this case the redirection mechanism, the platform will respond in different ways:

- 302 with the redirection in case the given compact identifier was resolved successfully 
- 404 if the request was submitted on a valid compact identifier, but its prefix is not registered in [identifiers.org](https://identifiers.org)
- 400 if the submitted compact identifier for resolution is not valid, and this may be for a variety of reasons
I hope you find this information useful.
Kind Regards,
Manuel

mbdebian: All IDs will resolve, as long as they match the regular expression that defines the ID space for their associated prefix.
But, taking into account that, in a given namespace (prefix), there may be multiple resouces (a.k.a. providers) registered, it is possible that not all of them cover the whole ID space, so you may be redirected to a provider that returns (or should return) an error back for a given ID that is not covered.

docs/asciidoc/front_matter.adoc
mbdebian: [identifiers.org](https://identifiers.org) resolution services are available via HTTP and HTTPS, we don't enforce any of the mechanisms like we do with the [registry](https://registry.identifiers.org), which is only available via HTTPS, because the data in [identifiers.org](https://identifiers.org) resolution services is open and public, but the [registry](https://registry.identifiers.org) supports other activities that require of AAA mechanisms

docs/asciidoc/front_matter.adoc
MichaelLukowski: I think it would be fine to just have have DRS clients resolve prefixes through identifiers.org. I agree that having this switch statement of language would be difficult to support long term and it would give greater flexibility to allow DRS implementers to choose which prefix they would want.

docs/asciidoc/front_matter.adoc
MichaelLukowski: agreed, will change and make new commit

docs/asciidoc/front_matter.adoc
MichaelLukowski: Thanks for catching this, should only be there once. I will fix and push commit.

dglazer:
FYI -- see #315 for a fleshed out version of https://github.com/ga4gh/data-repository-service-schemas/pull/312#issuecomment-597211476, intended to provide a simplified version of the approaches proposed here and in #313

docs/asciidoc/front_matter.adoc
ianfore: Pull request  #316 illustrates my point above that a drs: prefix itself may not be necessary. The use of identifiers.org and n2t.net as prefix registries and resolvers would do it

- Providing the required prefix functionality identified in this pull request
- Does so via an existing spec & implementation without the need for a new mechanism specific to DRS

MichaelLukowski:
We will be closing this PR

We're happy with the approach in this feature branch: https://ga4gh.github.io/data-repository-service-schemas/preview/feature/issue-306-unified-guid-pr/docs/#_overview

Issue #306, opened 2019-11-14
Persistent Identifiers and Prefixes


MichaelLukowski:
Right now the DRS endpoints only support simple object matching like

`drs://example.com/{object_id}`

As a driver project we have identified potential problems with endpoints using only using the server name and object id. We have found that working with funding from others and working with multiple large groups that URLs change for a variety of reasons. 

As it stands now a simple URL change from `example.com` to `foobar.com` would result in all of the previous DRS endpoints being broken. This is a problem as DRS cannot be used as a persistent identifier, thus researchers using these data sets cannot reliably cite these objects in their publications and research.

The solution we would like to propose would be a simple prefix in the URL to allow for an existing resolution service, like numbers to things (n2t.net) or identifiers.org, to resolve the endpoint in the event that a request returns a 404. The new endpoint would look something like 

`drs://example.com/{prefix}/{object_id}`

This change would allow DRS to align with ARKs (https://en.wikipedia.org/wiki/Archival_Resource_Key) and DOIs (https://en.wikipedia.org/wiki/Digital_object_identifier)

dglazer:
Michael, thank you for starting an issue to track this discussion. I'm adding a link to the [discussion document](https://docs.google.com/document/d/1E2dRQeGEtE1koBsYQTj6hadmENQABJHoLXGY9vFGwZ8/edit#heading=h.v4n2syv7eyfc) we've been using, in case others find that background, and the open questions it raises, useful context.

Two comments on the issue itself:
- this may be pedantic, but I disagree that "_DRS endpoints only support simple object matching_" -- I'd say they do even less than that, and instead are simple REST methods that only do exactly what they're asked to do. The `drs://` syntax you list isn't part of any DRS endpoint; it's a recommended text interchange format that's parsed by DRS clients, and tells them what DRS server to connect to, and what request to send to what endpoint. (In exactly the same way that `http://` syntax is parsed by HTTP clients and tells them what server to connect to, and what request to send to what endpoint.)
- as discussed in the document, directionally I'm all for the idea of adding a prefix of some sort that clients can use to find the right DRS server, and then ask it how to fetch content described by a server-relative id. The trick is in the details of how the prefix is assigned, specified, and resolved. I look forward to seeing a specific proposal.

pgrosu:
[...]

PR #315, opened 2020-03-21
Simple URI redirection (addresses #306)

dglazer:
This PR fleshes out the proposal in https://github.com/ga4gh/data-repository-service-schemas/pull/312#issuecomment-597211476. It is intended to satisfy the desire for more persistent URIs, using a simplified version of the approaches proposed in #312 and #313.

See https://ga4gh.github.io/data-repository-service-schemas/preview/feature/issue-306-simple-uri-redirection/docs/ to preview the human-readable documentation -- the only change is a new section 3.2.1 on URI Redirection.

### DIFF
--- a/docs/asciidoc/front_matter.adoc
+++ b/docs/asciidoc/front_matter.adoc
@@ -23,7 +23,7 @@ For convenience, including when passing content references to a WES server, we d
 For example, if a WES server was asked to process `drs://drs.example.org/314159`, it would know that it could issue a GET request to `https://drs.example.org/ga4gh/drs/v1/objects/314159` to learn how to fetch that object.
 
 ==== URI Redirection
-One potential concern with this URI syntax is that it depends on the server name (e.g. `drs.example.org`) remaining valid. In the rare event where a DRS server operator needs to preserve the validity of published resources, but can't preserved the validity of their DNS name, we define a standard mechanism for DRS server operators to redirect DRS clients to a new physical server.
+One potential concern with this URI syntax is that it depends on the server name (e.g. `drs.example.org`) remaining valid. In the rare event where a DRS server operator needs to preserve the validity of published resources, but can't preserve the validity of their DNS name, we define a standard mechanism for DRS server operators to redirect DRS clients to a new physical server.
 
 If a *DRS client* tries to fetch content, and gets an error indicating that the DRS server in the URI doesn't exist, the client should:
 
--- a/docs/asciidoc/front_matter.adoc
+++ b/docs/asciidoc/front_matter.adoc
@@ -22,6 +22,20 @@ For convenience, including when passing content references to a WES server, we d
 
 For example, if a WES server was asked to process `drs://drs.example.org/314159`, it would know that it could issue a GET request to `https://drs.example.org/ga4gh/drs/v1/objects/314159` to learn how to fetch that object.
 
+==== URI Redirection
+One potential concern with this URI syntax is that it depends on the server name (e.g. `drs.example.org`) remaining valid. In the rare event where a DRS server operator needs to preserve the validity of published resources, but can't preserved the validity of their DNS name, we define a standard mechanism for DRS server operators to redirect DRS clients to a new physical server.
+
+If a *DRS client* tries to fetch content, and gets an error indicating that the DRS server in the URI doesn't exist, the client should:
+
+* call `identifiers.org` with `drs:servername`, which will 302 them to a new cacheable servername (e.g. `drs:obsolete.org` will redirect to `newhotness.org`.)
+* use the new servername to resolve the updated DRS URI normally (e.g. if they're redirected to `newhotness.org`, they should act is if the URI were `drs://newhotness.org/314159` and issue a GET request to `https://newhotness.org/ga4gh/drs/v1/objects/314159`)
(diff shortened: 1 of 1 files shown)


dglazer:
Thanks for digging into the details @briandoconnor . A few quick reactions/questions:

1. Are either of the issues you mention (accession syntax, security) easier if we use one of the other popular resolver services? There's nothing special about identifiers.org if so.

2. For accession syntax, can we just use an escaping convention, and (for instance) tell clients to look up `bdcatalyst.dataguids.org` by passing `drs:bdcatalyst%datauids%org` to identifiers.org? It's a little funny looking, but that doesn't matter as long as we clearly document the behavior for servers and clients.

3. For security, isn't that a problem for *any* proposal that depends on identifiers.org? Or is there something about the other proposals that avoids the risk?

briandoconnor:
Hey David,

Let me take a look at each...

1. This is where I'm limited in my expertise, I don't really follow the GUID/compact identifier space closely enough to see if there's an alternative that makes sense.  Certainly, identifiers.org seems popular and I really like that you can use it with any compact identifier (versus other choices where we would have to tell people to use a particular GUID type).  It's very nice in that you can use it as both a registry of resolvers (which most DRS clients would use it for) and a proxy for resolving too.  We could go directly to Arks, Minids, Data Guids, or DOIs but then I think we lose the ability to support an abstraction on top of GUID types.  I think someone else needs to jump in here and comment on alternatives, though.  

2. That would be a totally workable fix if it was allowed.  However, like DRS IDs, namespaces in identifiers.org do not support encoding unfortunately.  You can just use [a-z] and "."

For your reference here's the complete list of namespaces (and provider codes): https://n2t.net/e/cdl_ebi_prefixes.yaml

3. Ya, this lookup security hole affects any proposal that uses GUIDs or has a fallback to looking up via GUID resolution.  That's why, the more I think about it, the more I'm favoring what I original suggesting: 1) one URI scheme that only uses DNS which is much harder to hijack and another that 2) uses GUIDs but is more fuzzy and open to man in the middle attacks but still is "nice" in that you have the longevity and flexibility of GUIDs.

Option 1 is good if you really want to control exactly the server the user is authorizing with to access data and I suspect folks serving controlled data, like NIH dbGaP/SRA for example, would want to take this approach.  The GUID approach option 2 is nice if your concern is more about the longevity of the data objects that you're creating and you care less about security e.g. it's open, reference data and you want it to stay available over the long run.

I think concern 3 is really opening up a larger question about how passports are used for services like WES and DRS.  If the passport is all powerful, how/when do you pass it to API endpoints whether they are WES, DRS, or something else?  It's one thing when you have a 1:1 relationship between bearer tokens and the services that use them.  But in this world of passports, how do you know it's safe to pass it around?  This is something that I need to understand better but the core issue of trust is coming up in the GUID example since it really does introduce a third party into the equation with fairly lax standards about who can register resolvers/mirrored resolvers.

dglazer:
1. Re other resolver services: got it. If that ends up being a fix we can see what Bob and others propose; let's keep talking through the specific blockers on identifiers.org for now.

2. Re valid accession syntax -- where is the "one dot" limitation documented? If I go to identiiers.org and try to resolve a `doi:` reference (see below), it tells me that I can use accessions with any number of dots, as long as they start with something like 12.3456. That suggests that we could define a workable regex for `drs:` (although I haven't found anything definitive).
image

3. Re malicious attacks -- hmm; I really don't feel good about publishing a standard that says "you have to choose between longevity and security". If resolver services are fundamentally insecure, I don't think we can use them. Open to discussion, but that feels fatal.

4. One new thought about using identifiers.org in a slightly less insecure way -- if I read it correctly, we can specify an explicit `provider` which would keep pirates at bay. So for example we could tell clients to look up `ga4gh/drs:dataguids.org`, and since we specified `ga4gh` explicitly, it would only look for those entries, and nobody could create malicious mirrors. (That only works if there's some way to ensure only legit people can make updates to identifiers.org, of course -- I'm not sure how that's protected.)

MichaelLukowski:
Hey David and Brian,

I think identifiers.org is going to have to be the way that we will have to move. It is a global registry and is linked with other resources like n2t for multiple places to resolve the same prefix.

Prefixes are registered and then there is a regular expression that is created that is part of the id of the prefix.

As for the security concerns. Once a prefix is registered then I don't know if you can cancel that and have a different (potentially malicious) organization register the same prefix. I am assuming that there is some sort of accountability that is investigated on identifiers.org part before prefixes are resolved.

There are a lot of checks that identifers.org has in place before someone is even granted a prefix that can be resolved.

dglazer:
@MichaelLukowski -- that would be great if identifiers.org works, since I agree it has a lot of advantages. We just need someone authoritative to address the concerns that Brian raised about id syntax and security. I'm optimistic there are good answers along the lines you suggest, but I didn't find them in 20 minutes of reading the docs.

dglazer:
I'm closing this PR -- as  the discussion has evolved, it's no longer the best option. I believe #318 is the closest to viable, so will focus on improving it.

I still like some features of this PR more than the alternatives I've seen -- in particular the simplicity, and the ability to use meta-resolvers as a fallback to DNS. Both of those features have challenges, though, so we may not be able to preserve them

PR #316, opened 2020-03-22
Simplification based on CURIEs - (addresses #306) 

ianfore:
This PR uses the proposal in #315. Provides support for prefixes entirely through CURIES where the prefixing requested in #306 becomes the CURIE prefix. 
Uses identifiers.org and n2t.net as
- registry of prefixes
- redirection to appropriate DRS server
- handling changes to DRS server URLs

### DIFF
--- a/docs/asciidoc/front_matter.adoc
+++ b/docs/asciidoc/front_matter.adoc
@@ -8,6 +8,8 @@ The primary functionality of DRS is to map a logical ID to a means for physicall
 
 === DRS IDs
 
+DRS id refers to the <id> in the CURIE syntax described below.
+
 Each implementation of DRS can choose its own id scheme, as long as it follows these guidelines:
 
 * DRS IDs are strings made up of uppercase and lowercase letters, decimal digits, hypen, period, underscore and tilde [A-Za-z0-9.-_~]. See https://tools.ietf.org/html/rfc3986#section-2.3[RFC 3986 § 2.3].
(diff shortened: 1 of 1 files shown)


ianfore:
A problem is that some prefixes already resolve to an html landing page. In the DRS case a json response is expected.

It would help to know if identifiers/n2t have implemented mechanisms whereby one can choose whether a requester can specify
a) whether they want machine readable content returned instead of a landing page
or
b) whether they could specify they wanted a particular kind of content

That said registering a prefix which returns a machine readable representation does work with the existing set up. In one case I registered the cadsr prefix which returns xml
See here how identifiers.org passes on resolution of [cadsr:5750671](http://identifiers.org/cadsr:5750671) for example.

I guess my backstop on this problem was that a program expecting DRS content would look at the response and handle it appropriately. It might not happen often because, the convention would be that certain prefixes deliver a DRS response.  dg: would be one, but as noted elsewhere not the only one. Convention might get criticized for lack of robustness, but may be good enough in the circumstance.

ianfore:
Illustration attached as to how I imagined a prefix wold be registered for dg: under this proposal. 

Some of the fields get truncated by the forms UI, but I think enough is visible to convey understanding.

[Request an identifier prefix.pdf](https://github.com/ga4gh/data-repository-service-schemas/files/4406031/Request.an.identifier.prefix.pdf)

ianfore:
Closing and deferring to https://github.com/ga4gh/data-repository-service-schemas/pull/318 for comprehensive solution.

PR #318, opened 2020-04-23
Feature/issue 306 unified guid pr

briandoconnor:
## TL;DR

Compact identifier-based DRS URIs would be super helpful since they replace the hostname in the URI scheme we support now with a prefix that can easily be looked up on identifiers.org/n2t.net and mapped to the current DRS server for that ID. This improves the longevity of DRS URIs and increases their FAIR-ness.  The goal of this PR is to provide a unified approach to using compact identifiers in DRS URIs in addition to the hostname-base DRS URIs already in the spec.  These compact identifier-based URIs looks like `drs://prefix:accession`, for example `drs://doi:10.5072/FK2805660V`. Compact identifier-based DRS URIs can work with existing providers (DOIs, ARKs, Data GUIDs) or a DRS server might choose to create their own compact identifier type.  It makes the URI -> URL resolution process for DRS clients more complex, this is why it  wouldn't replace our current hostname-based DRS URIs, just supplement it.

## Proposal Highlights

* A single DRS URI scheme "drs://"
* Two approaches to URI: hostname-based (what we have now) and compact-identifier based (new).
* Support for compact identifiers based on the CURIE format used by identifiers.org/n2t.net
* Rules for a DRS client on how to tell these two URI styles apart and how to translate each to a valid DRS HTTPS GET request.
* Robustly support any compact identifier registered on identifiers.org/n2t.net (both new compact identifiers and existing like ARKs, DOIs, Data GUIDs etc)
* Retain the ability to use hostname-based DRS URIs for systems with security requirements that will not allow for the use of third party resolver registries.
* Specifically avoid using identifiers.org/n2t.net as a fallback for hostname resolution per the recommendation by those groups.  See the [security discussion](https://docs.google.com/document/d/1vd75WUNO8lv_i69huap2UX9lsEq3WI5tJwTqwVZ7kN8/edit#) for more info.

## Driver Vote Summary

| Driver Project  | Vote  |
| -------------- | ------ |
| AGHA | 0 |
| HCA DCP | +1 |
| NCI Cancer Research Data Commons | +1 |
| TOPMed | +1 |
| NCI GDC | +1 |
| CanDIG | 0 |
| ELIXIR Beacon | 0 |
| EMBL-EBI (Implementer) | +1 |
| Seven Bridges (Implementer) | +1 |
| NCBI/NLM/NIH | +1 |

## Updated Specification Document

You can see the specification updates rendered in our documentation here, this is much easier to follow than looking at a markdown diff:

https://ga4gh.github.io/data-repository-service-schemas/preview/feature/issue-306-unified-guid-pr/docs/

In particular look at sections 3.2, 8, 9, and 10.

_If you're curious about comparing the rewritten version of this PR based on David G's comments to the original PR you can see the original here:_
[PR_318_pre_changes_from_David.zip](https://github.com/ga4gh/data-repository-service-schemas/files/4607190/PR_318_pre_changes_from_David.zip)

## The Details

*To be crystal clear, this PR is not a change to the [DRS API](https://ga4gh.github.io/data-repository-service-schemas/preview/release/drs-1.0.0/docs/) rather it's an addition to supported URI conventions for use with DRS clients.*  In the 1.0 release of DRS, we defined a simple URI scheme (drs://) that was combined with a DRS hostname and data object ID.  This simple URI could then be interpreted by DRS clients, using a simple rule, to the appropriate URL and subsequent GET request on a DRS server.  

A parallel approach to the DRS hostname-based URIs is to use compact identifiers instead of hostnames and object IDs in the URI.  [Compact identifier](https://www.biorxiv.org/content/10.1101/101279v3) are extremely frequently used in the research community and multiple Driver projects want to use them in place of the hostname-based DRS URIs we defined previously.  The advantage is projects can change hostnames (or even project names) and the globally unique compact identifiers aren't affected.  There also exists a standard service (identifiers.org/n2t.net) for resolving a compact identifier to resolvers, in this case providing a mapping from compact identifier to a DRS HTTPS URL.  Identifiers.org/n2t.net provide a third party service to map a compact-identifier to a concrete URL rather than the rule-based approach we used in hostname-based DRS URIs currently described in the spec.

*The goal of this PR is to provide a clear description of both compact identifier-based DRS URIs and hostname-based DRS URIs so that a DRS client can successfully resolve both types into fetchable DRS HTTPS URLs.*

An issuer of DRS URIs, therefore, needs to choose if they want the security/performance of hostname-based URIs or the flexibility and longevity of compact identifier-based URIs.  In truth, both might be used at different times by the same DRS server.  For example, hostname-based URIs used for immediate execution in a workflow system where performance and integrity of the data source are key priorities and compact identifier-based URIs used in a research paper where longevity is paramount.  *DRS clients _must_ support both.*

Much more research into the approach I present in this PR can be found in the document [20200322 - DRS and GUIDs v2](https://docs.google.com/document/d/13ZCcF08FzRkmwpLb9si-CDuYVOdB2CsgprmYfhB-Dt4/edit#heading=h.8rt50ql1ypu1)

More information on the security implications of using identifiers.org/n2t.net can be found [here](https://docs.google.com/document/d/1vd75WUNO8lv_i69huap2UX9lsEq3WI5tJwTqwVZ7kN8/edit#heading=h.8yf5b8pa109m)

## Consensus PR

standards

This PR is intended to harmonize 4 different PRs (see below) that have been opened by the community in an attempt to address using compact identifiers, for example [Data GUIDS](https://dataguids.org/), with the DRS URIs.  These PRs propose various formats for compact identifier-based DRS URIs and conventions/approaches to resolving them into valid DRS URLs.  

## Addressing Ideas from Open PRs

**These are all closed now**

I tried to leverage and merge ideas from each of the previous 4 PRs into this single PR.  I wasn't always able to use a particular PR's approach unmodified given that I'm trying to find consensus here.  But I tried to address what each PR was focused on:

* [PR#312](https://github.com/ga4gh/data-repository-service-schemas/pull/312):  PR#312 wanted to include a prefix in the DRS URI format but it was ambiguous how a client would know how to resolve these.  It also made changes to the DRS API schema itself.  **-> CLOSED:** the author has closed the PR since the consensus PR meets their needs
  * *Response* - I included prefix (and provider code) support in this PR to address this in a similar way.  But this PR proposes two forms of the DRS URI so that either hostnames or compact identifiers can be used in DRS URIs.  Compact identifiers aren't used as a fallback to DNS in this PR.  Instead, this PR makes it clear that identifiers.org/n2t.net are used to resolve DRS URIs using compact identifiers.  Clients won't need special knowledge about prefixes in order to work. This PR also makes no changes to the DRS API schema itself, it is purely focused on documentation changes.
* [PR#313](https://github.com/ga4gh/data-repository-service-schemas/pull/313): This PR is similar to #312 but, instead, uses an overloaded DRS URI.  It first tries to use the compact identifier prefix as a hostname and then falls back on resolving it as a compact identifier. **-> CLOSED:** the author has closed the PR since the consensus PR meets their needs
  * *Response* - this PR proposes, instead, to have a single URI scheme but a clear regex for compact identifier-based or hostname-based DRS URIs.  This prevents using DNS-resolution failures as part of the normal URI resolution process, which creates performance and security concerns.  The PR proposed here uses identifier.org/n2t.net for resolving prefixes from compact identifiers rather than listing them out as specific examples in the spec (ark: dg:) as PR#313 and PR#312 both do.
* [PR#315](https://github.com/ga4gh/data-repository-service-schemas/pull/315/): This PR uses a similar fallback strategy as PR#313 when a DNS-based DRS URI fails to resolve.  The idea is to register a common prefix on identifiers.org/n2t.net ("drs") and then resolve failed domain names (inaccessible-server.org) as compact identifiers like "drs:inaccessible-server.org.  Two drawbacks: 1) identifiers.org/n2t.net respond (in their lookup resolver role and not their accession redirect role) with a resolver record based on the prefix ("drs") not the accession ("inaccessible-server.org") so we can't use it to redirect to different DRS servers based on accessions.  If we want a "drs:" prefix with accessions per drs server we would need to have identifiers.org/n2t.net deploy a customized registry for us in parallel with the existing registry (a lot of work and parallel infrastructure)  2) the identifiers.org/n2t.net teams recommended against using their service as a fallback to DNS failures. **-> CLOSED:** the author has closed the PR since this approach isn't viable  
  * *Response* - this PR proposes not using a fallback mechanism with identifiers.org/n2t.net since that was not recommended.  Instead we recommend DRS URI issuers pick a style that's appropriate for their goals (hostname- or compact identifier-based) and they register their own prefixes on identifiers.org/n2t.net rather than trying to have GA4GH control the prefix.  *Identifiers.org/n2t.net's existing registry provides exactly the functionality we need and there's no benefit to setting up a parallel but different resolver registry just for DRS.*
* [PR#316](https://github.com/ga4gh/data-repository-service-schemas/pull/316/files): the core of this PR is to use the CURIE format of identifiers.org/n2t.net to directly leverage those services in the way they were designed.  It assumes all DRS resolution will happen through identifiers.org/n2t.net which gives a natural place to make updates to DRS server hostnames however consumers of DRS CURIEs may have difficulties in identifying these as pointers to DRS objects without first resolving them using identifiers.org/n2t.net and guessing based on their URL pattern. **-> CLOSED:** the author has closed the PR since the consensus PR meets their needs
  * *Response* - this PR proposes using the exact same CURIE format for compact identifier-based DRS URIs but does prefix with the drs:// scheme (which means it is not a CURIE) in order to provide a clear hint to systems that they can expect to get a DRS URL once the compact identifier is resolved.  By using the same CURIE format of prefix:accession (rather than the custom format from PR#312 and PR#313) I'm hoping to work more cleanly with identifiers.org/n2t.net and enable _any_ compact identifier to be used.  Also, I used the clear and standard identifiers.org/n2t.net resolution process proposed here as well, rather than adapting these services as a backup to DNS resolution (PR#315).


abaumann:
👍 this seems to me to be what's need from the Terra side for support for both guids and hostnames

docs/asciidoc/back_matter.adoc
```diff
@@ -39,3 +37,111 @@ This spec defines a standard **Data Repository Service (DRS) API** (“the yello
...
+
+See the documentation on https://n2t.net/e/compact_ids.html[n2t.net] and https://docs.identifiers.org/[identifiers.org] for much more information on the compact identifiers used there and details about the resolution process.
+
+.Note: DRS Client Compact Identifier-Based URI Resolution Process
+****
+A DRS client identifies the a DRS URI compact identifier components using the first occurance of "/" (optional) and ":" characters. These are not allowed inside the provider_code (optional) or the namespace. The ":" character is not allowed in a Hostname-based DRS URI, providing a convenient mechanism to differentiate them. Once the provider_code (optional) and namespace are extracted from a DRS compact identifier-based URI, a client can use services on identifiers.org to identify available resolvers.
```
jiaqi216, line 96: > A DRS client identifies the a DRS URI compact identifier components using the first occurance of "/" (optional) and ":" characters.

I'm a little bit lost on the first occurrence of `/` since it can be interpreted to be the `/` in `drs://`. So a DRS client knows that it is a compact identifier-based URI by optional `/` after the provider code and the `:`. It might be clearer to just emphasize that the indication of `:` in the DRS ID will signal a compact identifier-based URI if it will always exist

docs/asciidoc/back_matter.adoc
```diff
@@ -39,3 +37,111 @@ This spec defines a standard **Data Repository Service (DRS) API** (“the yello
...
+
+Still, to make compact identifiers work in DRS URIs we leverage the CURIE format used by identifiers.org/n2t.net.  Compact identifiers have the form:
+
+    prefix:accession
+
+The prefix can be divided into a `provider_code` (optional) and `namespace`.  The `accession` here is an Ark, DOI, Data GUID, or another issuers's local ID for the object being pointed to:
+
+    [provider_code/]namespace:accession
+
+Both the `provider_code` and `namespace` disallow spaces or punctuation, only lowercase alphanumerical characters, underscores and dots are allowed.
```
jiaqi216, line 78: I think they allow for both lower and uppercase letters so`[A-Za-z0-9._]`

docs/asciidoc/front_matter.adoc
```diff
@@ -1,8 +1,13 @@
-== Introduction
+////
+TODO:
+* I need to clear up the section showing the GET requests, clarify the "ID", and also probably mention these API calls are examples and we make not guarantees that identifiers.org/n2t.net won't change.
+* do we want to add a self_url to the JSON response from DRS objects?  This would make it much, much easier to construct the access method rather than using the hostname-based URI which requires manipulation and parsing.
```
jiaqi216, line 4: Yeah, I think that could be useful. At least as an optional field so DRS servers can opt out if they want to.

MichaelLukowski:
This will help U Chicago / Gen3 greatly with support for guids and hostname resolution

docs/asciidoc/front_matter.adoc
```diff
@@ -11,16 +16,56 @@ The primary functionality of DRS is to map a logical ID to a means for physicall
 Each implementation of DRS can choose its own id scheme, as long as it follows these guidelines:
 
 * DRS IDs are strings made up of uppercase and lowercase letters, decimal digits, hypen, period, underscore and tilde [A-Za-z0-9.-_~]. See https://tools.ietf.org/html/rfc3986#section-2.3[RFC 3986 § 2.3].
-* Note to server implementors: internal IDs can contain other characters, but they MUST be encoded into valid DRS IDs whenever exposed by the API.   
+* DRS IDs can contain other characters, but they MUST be encoded into valid DRS IDs whenever exposed by the API.  This is because non-encoded IDs may interfere with the interpretation of the `objects/{id}/access` endpoint.  To overcome this limitation use percent-encoding of the ID, see https://tools.ietf.org/html/rfc3986#section-2.4[RFC 3986 § 2.4]
```
jiaqi216, line 19: A small clarification on language - would this be clearer if it says "MUST Be encoded into valid DRS IDs whenever exposed _to_ the API" instead of _by_? Since, in the event of other characters, clients will need to make sure proper percent-encoding is handled when passing the DRS ID to the `objects/{id}/access` endpoint in order to get a valid response.

openapi/data_repository_service.swagger.yaml
```diff
@@ -203,8 +203,8 @@ definitions:
       self_uri:
         type: string
         description: |-
-          A drs:// URI, as defined in the DRS documentation, that tells clients how to access this object.
-          The intent of this field is to make DRS objects self-contained, and therefore easier for clients to store and pass around.
+          A drs:// hostname-based URI, as defined in the DRS documentation, that tells clients how to access this object.
+          The intent of this field is to make DRS objects self-contained, and therefore easier for clients to store and pass around.  For example, if you arrive at this DRS JSON by resolving a compact identifier-based DRS URI, the `self_uri` presents you with a hostname and properly encoded DRS ID for use in subsequent `access` endpoint calls.
```
jiaqi216, line 207: If I'm reading this correctly, the self_uri is always a hostname-based uri?

docs/asciidoc/back_matter.adoc
briandoconnor: Yep, thanks, I made a note

docs/asciidoc/back_matter.adoc
briandoconnor: The spec has been reorganized but this is what the text currently says:

1) The client parses the string to extract the prefix of `drs.42` and the accession of `314159`, using the first occurrence of a colon (":") character after the initial `drs://` as a delimiter. (The colon character is not allowed in a Hostname-based DRS URI, making it easy to tell them apart.)

I think that better matches what you're proposing here.

docs/asciidoc/front_matter.adoc
briandoconnor: yep, I'll file a bug for this one so we don't loose track of the idea for a future version

briandoconnor: See https://github.com/ga4gh/data-repository-service-schemas/issues/320

docs/asciidoc/front_matter.adoc
briandoconnor: Thanks, I changed this to "whenever they are used in API calls"

openapi/data_repository_service.swagger.yaml
briandoconnor: Right now yes. I opened a new issue to explore the possibility of including a self_url in the response since that would make it even easier to construct the "access" request.

briandoconnor:
Today I pushed changes to the spec proposed by David G and Ian F.  I incorporated those into the current specification which remains unchanged in technical detail even if though the description has been heavily revised.  I included a supplement document outside of the spec that includes the additional examples of resolving compact identifier-based DRS URIs that have been removed from the revised spec.

These revision point to a few questions for Drivers that are going to vote on this:

* do we support compact identifiers based on the format used by identifiers.org/n2t.net or do we support a subset of the format (only namespaces)?
* do we mandate that every DRS server use a namespace starting with `drs.`?

There are +/- for each, let's discuss here but I recommend settling the issue based on Drivers needs and doing so quickly since nobody is benefiting from delaying on DRS 1.1.

Next step is to identify the voting Driver projects, Rishi has done a survey and we'll discuss this in the next Cloud call.

docs/asciidoc/front_matter.adoc
```diff
@@ -11,16 +20,100 @@ The primary functionality of DRS is to map a logical ID to a means for physicall
...
+* DRS IDs can contain other characters, but they MUST be encoded into valid DRS IDs whenever they are used in API calls.  This is because non-encoded IDs may interfere with the interpretation of the `objects/{id}/access` endpoint.  To overcome this limitation use percent-encoding of the ID, see https://tools.ietf.org/html/rfc3986#section-2.4[RFC 3986 § 2.4]
 * One DRS ID MUST always return the same object data (or, in the case of a collection, the same set of objects). This constraint aids with reproducibility.
-* DRS v1 does NOT support semantics around multiple versions of an object. (For example, there’s no notion of “get latest version” or “list all versions”.) Individual implementation MAY choose an ID scheme that includes version hints.
 * DRS implementations MAY have more than one ID that maps to the same object.
+* DRS version 1.x does NOT support semantics around multiple versions of an object. (For example, there’s no notion of “get latest version” or “list all versions”.) Individual implementation MAY choose an ID scheme that includes version hints.
```
dglazer, line 26: s/implementation/implementations/

docs/asciidoc/front_matter.adoc
```diff
@@ -11,16 +20,100 @@ The primary functionality of DRS is to map a logical ID to a means for physicall
...
+
+TIP: See <<_appendix_background_notes_on_drs_uris>> for more information on our design motivations for DRS URIs.
+
+==== Hostname-based DRS URIs
+
+Hostname-based DRS URIs are simpler than compact identifier-based URIs.  They contain the DRS server name and the DRS ID only and can be converted directly into a fetchable URL based on a simple rule.  They take the form:
+
+    drs://<hostname>/<id>
+
+DRS URIs of this form mean “you can fetch the content with DRS id <id> from the DRS server at <hostname> ".
```
dglazer, line 43: Minor: there's an extra trailing space inside the quotation marks.
More minor: did you try italicizing the quoted text? Either way works for me.

docs/asciidoc/front_matter.adoc
```diff
@@ -11,16 +20,100 @@ The primary functionality of DRS is to map a logical ID to a means for physicall
...
+|===
+| |Hostname-based |Compact Identifier-based
+
+|URI Durability
+|URIs are valid for as long as the server operator maintains ownership of the published DNS address. (They can of course point that address at different physical serving infrastructure as often as they’d like.)
+|URIs are valid for as long as the server operator maintains ownership of the published compact identifier resolver namespace. (They also depend on the meta-resolvers like identifiers.org/n2t.net remaining operational, which is intended to be essentially forever.)
+
+|Client Efficiency
+|URIs require minimal client logic, and no network requests, to resolve.
+|Citation practice is consistent with publisher instructions. URIs require small client logic, and 1-2 cacheable network requests, to resolve.
```
dglazer, line 107: Per earlier discussion, I'm not convinced that either URI style is more consistent with publisher instructions. (The only publisher instructions I saw were related to durability, which is covered   above.)

If we do track down publisher instructions that distinguish between the two styles, I suggest moving this sentence to the end of the bullet (to keep it more parallel to the lefthand column.)

docs/asciidoc/front_matter.adoc
```diff
@@ -11,16 +20,100 @@ The primary functionality of DRS is to map a logical ID to a means for physicall
...
+
+|Security
+|Servers have full control over their own security practices.
+|Server operators, in addition to maintaining their own security practices, should confirm they are comfortable with the resolver registry security practices, including protection against denial of service and namespace-hijacking attacks. (See the <<_appendix_compact_identifier_based_uris>> for more information on resolver registry security.)
 
-For example, if a WES server was asked to process `drs://drs.example.org/314159`, it would know that it could issue a GET request to `https://drs.example.org/ga4gh/drs/v1/objects/314159` to learn how to fetch that object.
+|https://www.nature.com/articles/sdata201618[FAIR-ness]
+|Provided the published DNS address is valid.
+|Same assuming the resolver registry namespace is valid. Slightly increased find-ability given the extra metadata available in the resolver registry (contact, organization, etc).
```
dglazer, line 115: I don't understand this row -- URI durability is definitely important for FAIRness, but that's already covered above. And the extra metadata is only at the level of the resource as a whole (vs. individual objects), which I believe is already covered by /service-info in a way that works for all URIs.

docs/asciidoc/more_background_on_compact_identifiers_content.adoc
```diff
@@ -0,0 +1,145 @@
+== About
+
+This document contains more examples of resolving compact identifier-based DRS URIs than we could fit in the DRS specification or appendix.  It's provided here for your reference as a supplement to the specification.
+
+== Hostname-Based URIs
```
dglazer, line 5: Did you mean to include this info on Hostname-Based URIs in the new document? (I think there may be other unintended content, but will wait until you've had a chance to respond before looking more closely.)

docs/asciidoc/more_background_on_compact_identifiers_content.adoc
```diff
@@ -0,0 +1,145 @@
+== About
```
dglazer, line 1: Format question -- did you think about just putting this into a .md file instead, since it doesn't need to be tightly coupled to the generated docs? I expect it will be a lot easier to maintain, since it avoids the whole build process. (And if I understand James' work, we'll be switching everything from .adoc to .md anyway..)

docs/asciidoc/back_matter.adoc
```diff
@@ -39,3 +37,90 @@ This spec defines a standard **Data Repository Service (DRS) API** (“the yello
...
+* We support compact identifier-based URIs, and the meta-resolver services of identifiers.org and n2t.net (Name-to-Thing), because of the wide adoption of compact identifiers in the research community. as detailed by https://doi.org/10.1038/sdata.2018.29[Wimalaratne et al (2018)] in "Uniform resolution of compact identifiers for biomedical data."
+
+== Appendix: Compact Identifier-Based URIs
+
+.Note: Identifiers.org/n2t.net Changes
+****
+The examples below show the current API interactions with https://n2t.net/e/compact_ids.html[n2t.net] and https://docs.identifiers.org/[identifiers.org] and these may change over time.  Please refer to the documentation from each site for the most up-to-date information.  We will make best efforts to keep the DRS specification current but DRS clients MUST maintain their ability to use either the identifiers.org or n2t.net APIs to resolve compact identifier-based DRS URIs.
+****
+
+=== Registering a DRS Server on a Meta-Resolver
```
dglazer, line 58: There's a lot of content below here that looks redundant with the new separate doc on CI's -- did you mean to delete it from this file?

dglazer: Let me know once this is cleaned up and I'll review the rest of the Appendix more closely.

docs/asciidoc/back_matter.adoc
```diff
@@ -39,3 +37,90 @@ This spec defines a standard **Data Repository Service (DRS) API** (“the yello
...
+* We support hostname-based URIs because of their simplicity and efficiency for server and client implementers.
+* We support compact identifier-based URIs, and the meta-resolver services of identifiers.org and n2t.net (Name-to-Thing), because of the wide adoption of compact identifiers in the research community. as detailed by https://doi.org/10.1038/sdata.2018.29[Wimalaratne et al (2018)] in "Uniform resolution of compact identifiers for biomedical data."
+
+== Appendix: Compact Identifier-Based URIs
+
+.Note: Identifiers.org/n2t.net Changes
+****
+The examples below show the current API interactions with https://n2t.net/e/compact_ids.html[n2t.net] and https://docs.identifiers.org/[identifiers.org] and these may change over time.  Please refer to the documentation from each site for the most up-to-date information.  We will make best efforts to keep the DRS specification current but DRS clients MUST maintain their ability to use either the identifiers.org or n2t.net APIs to resolve compact identifier-based DRS URIs.
```
dglazer, line 55: minor: replace "and these" with ", which"

docs/asciidoc/front_matter.adoc
briandoconnor: good catch

docs/asciidoc/front_matter.adoc
briandoconnor: Done

docs/asciidoc/front_matter.adoc
briandoconnor: OK, I moved it down.  A quick search picks up this article in nature about recommendations for using compact identifiers in publications but maybe Ian has examples of publishers including compact identifiers as a requirement in publications?  See https://www.nature.com/articles/sdata2018259

docs/asciidoc/front_matter.adoc
briandoconnor: I added because I didn't see findable, accessible, interoperable, and reproducible specifically called out in the concepts durability, security, and efficiency.  I think that's a great point about 1) service-info and 2) the service registry GA4GH is working on.  When those are both in place then I think hostname and compact identifier are equal from a FAIR perspective. That's just my opinion.

docs/asciidoc/more_background_on_compact_identifiers_content.adoc
briandoconnor: Yep, I removed hostnames from this doc and just added a note that accessions from compact identifiers may need encoding before being used as a DRS ID in DRS API calls given the characters allowed in DRS IDs.

docs/asciidoc/more_background_on_compact_identifiers_content.adoc
briandoconnor: Happy to revise when the new build system is in place.

docs/asciidoc/back_matter.adoc
briandoconnor: Done!  I removed redundancy that I saw between the spec doc and the background doc

docs/asciidoc/back_matter.adoc
briandoconnor: done

briandoconnor:
At this point I addressed the comments David and Jiaqi made.

docs/asciidoc/more_background_on_compact_identifiers_content.adoc
dglazer: Okay -- you can use .md now and it will be future-proof (and will keep the builds simpler in the meantime), but agree it's not at all pressing.

docs/asciidoc/front_matter.adoc
dglazer: Thanks for moving it down.

I've seen that article -- I think it's the same one Ian pointed out earlier. Two concerns:
- as you say those aren't actually _publisher instructions_, but rather a suggestion in a paper
- neither style is _consistent_ with the suggestion, since our URI's are not actually compact identifiers

Given that, I don't think it's accurate to say either style is consistent with publisher instructions.

briandoconnor:
## Driver Vote

* Driver Project: AGHA
* Vote: 0

Note: Rishi indicated this vote in the [spreadsheet](https://docs.google.com/spreadsheets/d/1xGwr5k6zduQI0_xne8s9t6JvbUBzrVPBd8QGCSbHT1k/edit#gid=169659378) on 20200511

docs/asciidoc/front_matter.adoc
dglazer: I still think we should delete this row -- I disagree that CI-based URIs are FAIRer, and I don't want to imply that anyone who chooses to use hostname-based URIs, is being less FAIR. (Re Findability -- I think of that as being about individual objects, which is identical between the two styles.)

docs/asciidoc/front_matter.adoc
dglazer: Now that the bug is filed, please remove this comment from the spec.

docs/asciidoc/front_matter.adoc
```diff
@@ -1,8 +1,17 @@
+////
+TODO:
+* do we want to add a self_url to the JSON response from DRS objects?  This would make it much, much easier to construct the access method rather than using the hostname-based URI which requires manipulation and parsing.
+* comments from David to note in the PR conversation
+  - removing provider_code
+	- mandating all DRS servers to register new prefixes
+	- saying 'drs.' should be in the prefixes
+////
```
dglazer, line 8: Please remove from the spec. (I'm not sure there are still relevant, but if so we should track them elsewhere.)

docs/asciidoc/front_matter.adoc
```diff
@@ -11,16 +20,102 @@ The primary functionality of DRS is to map a logical ID to a means for physicall
...
+|===
+| |Hostname-based |Compact Identifier-based
+
+|URI Durability
+|URIs are valid for as long as the server operator maintains ownership of the published DNS address. (They can of course point that address at different physical serving infrastructure as often as they’d like.)
+|URIs are valid for as long as the server operator maintains ownership of the published compact identifier resolver namespace. (They also depend on the meta-resolvers like identifiers.org/n2t.net remaining operational, which is intended to be essentially forever.)
+
+|Client Efficiency
+|URIs require minimal client logic, and no network requests, to resolve.
+|URIs require small client logic, and 1-2 cacheable network requests, to resolve.  Citation practice is consistent with publisher instructions.
```
dglazer, line 109: Please delete  "Citation practice is consistent with publisher instructions."  As discussed, the only reference we've seen is a vague suggestion, and one that neither URI style conforms to.

docs/asciidoc/more_background_on_compact_identifiers_content.adoc
```diff
@@ -0,0 +1,116 @@
...
+
+    GET https://drs.example.org/ga4gh/drs/v1/objects/10.5072%2FFK2805660V
+
+IDs in DRS hostname-based URIs/URLs are always percent-encoded to eliminate ambiguity even though the DRS compact identifier-based URIs and the identifier.orgs API do not percent-encode accessions.  This was done in order to 1) follow the CURIE conventions of identifiers.org/n2t.net for compact identifier-based DRS URIs and 2) to aid in readability for users who understand they are working with compact identifiers. **The general rule of thumb, when using a compact identifier accession as a DRS ID in a DRS API call, make sure to percent-encode it.  An easy way for a DRS client to handle this is to get the initial DRS object JSON response from whatever redirects the compact identifier resolves to, then look for the `self_uri` in the JSON, which will give you the correctly percent-encoded DRS ID for subsequent DRS API calls such as the `access` method.**
+
+
+== Example DRS Client Compact Identifier-Based URI Resolution Process - Registering a new Compact Identifier for Your DRS Server
```
dglazer, line 72: Let's move this section out of the spec and into the "more background" doc -- it will remove a lot of clutter, is only relevant to a minority of our users, and is the most likely to get out of date as other systems change.

docs/asciidoc/back_matter.adoc
```diff
@@ -39,3 +37,94 @@ This spec defines a standard **Data Repository Service (DRS) API** (“the yello
...
+
+(ii) The client extracts the namespace id (in this example 1234), and uses it to make a second GET request to identifiers.org to find information about the namespace:
+
+  GET https://registry.api.identifiers.org/restApi/resources/search/findAllByNamespaceId?id=1234
+
+This request returns a JSON structure including an urlPattern field, whose value is an URL pattern containing a `${id}` parameter, such as:
+
+  "urlPattern" : "https://drs.myexample.org/ga4gh/drs/v1/objects/{$id}"
+
+=== Calling the n2t.net API as a Client
```
dglazer, line 94: This should be indented as the second of two sub-sections.

docs/asciidoc/back_matter.adoc
```diff
@@ -39,3 +37,94 @@ This spec defines a standard **Data Repository Service (DRS) API** (“the yello
...
+
+image::prefix_register_1.png[]
+
+image::prefix_register_2.png[]
+
+=== Calling Meta-Resolver APIs for Compact Identifier-Based DRS URIs
+
+Clients resolving Compact Identifier-based URIs need to convert a prefix (e.g. “drs.42”) into an URL pattern. They can do so by calling either the identifiers.org or the n2t.net API, since the two meta-resolvers keep their mapping databases in sync.
+
+=== Calling the identifiers.org API as a Client
```
dglazer, line 72: This should be indented as the first of two sub-sections under "Calling Meta-Resolver APIs ..."

docs/asciidoc/back_matter.adoc
dglazer: Thanks -- resolving -- I'll comment on remaining issues separately.

docs/asciidoc/back_matter.adoc
```diff
@@ -39,3 +37,94 @@ This spec defines a standard **Data Repository Service (DRS) API** (“the yello
...
+=== Registering a DRS Server on a Meta-Resolver
+
+See the documentation on the https://n2t.net/e/compact_ids.html[n2t.net] and https://docs.identifiers.org/[identifiers.org] meta-resolvers for adding your own compact identifier type and registering your DRS server as a resolver. You can register new prefixes (or mirrors by adding resource provider codes) for free using a simple online form.  Keep in mind, while anyone can register prefixes, the identifiers.org/n2t.net sites do basic hand curation to verify new prefix and resource (provider code) requests.  See those sites for more details on their security practices.
+
+Starting with the prefix for our new compact identifier, let's register the namespace `mydrsprefix` on identifiers.org/n2t.net and use 5-digit numeric IDs as our accessions.  We will then link this to the DRS server at `https://mydrs.server.org/ga4gh/drs/v1/` by filling in the provider details.  Here's what that the registration for our new namespace looks like on https://registry.identifiers.org/prefixregistrationrequest[identifiers.org]:
+
+image::prefix_register_1.png[]
+
+image::prefix_register_2.png[]
+
```
dglazer, line 67: Let's move this section to the "More background info" doc -- it adds a lot of clutter, is only relevant to a minority of readers, and is the most likely part to get outdated if meta-resolvers change.

docs/asciidoc/back_matter.adoc
```diff
@@ -39,3 +37,94 @@ This spec defines a standard **Data Repository Service (DRS) API** (“the yello
...
+   redirect: https://drs.myexample.org/ga4gh/drs/v1/objects/$id
+
+=== Caching with Compact Identifiers
+
+Identifiers.org/n2t.net compact identifier resolver records do not change frequently.  This reality is  useful for caching resolver records and their URL patterns for performance reasons.  Builders of systems that use compact identifier-based DRS URIs should cache prefix resolver records from identifiers.org/n2t.net and occasionally refresh the records (such as every 24 hours).  This approach will reduce the burden on these community services since we anticipate many DRS URIs will be regularly resolved in workflow systems.  The implementation of the caching approach is up to system builders but we recommend at least caching the mapping of prefixes to URL patterns. System builders may also choose to cache the compact identifier accession to final DRS URL records as well, but that decision/design is at the discretion of the system builder.  Alternatively, system builders may decide to directly mirror the registries themselves, instructions are provided on the identifiers.org/n2t.net websites.
```
dglazer, line 108: What new information are we providing here, beyond the (important) guidance in the main text that clients SHOULD cache URL patterns? Unless I'm missing something, I suggest deleting this section, since as written it dilutes the main guidance. (I can provide more detailed comments on specific content, if you point out the key new messages you want to convey here.)

docs/asciidoc/front_matter.adoc
```diff
@@ -11,16 +20,102 @@ The primary functionality of DRS is to map a logical ID to a means for physicall
...
+
+===== For DRS Clients
+
+A DRS client parses the DRS URI compact identifier components to extract the prefix and the accession, and then uses meta-resolver APIs to locate the actual DRS server. For example, here are the client resolution steps if the URI is:
+
+    drs://drs.42:314159
+
+1) The client parses the string to extract the prefix of `drs.42` and the accession of `314159`, using the first occurrence of a colon (":") character after the initial `drs://` as a delimiter. (The colon character is not allowed in a Hostname-based DRS URI, making it easy to tell them apart.)
+
+2) The client makes API calls to a meta-resolver to look up the URL pattern for the namespace. (See <<_appendix_compact_identifier_based_uris>> for details.) The URL pattern is a string containing a `{$id}` parameter, such as:
```
dglazer, line 83: This should link directly to "Calling Meta-Resolver APIs for Compact Identifier-Based DRS URIs" if possible.

docs/asciidoc/more_background_on_compact_identifiers_content.adoc
dglazer: Looks like you want to leave it as .adoc for now, which is fine; resolving.

docs/sample_jsons/drsobject.24207d90-c5b8-4fac-8c07-6426521dfeac.json
```diff
@@ -0,0 +1,33 @@
+{
```
dglazer, line 1: Where is this new file referenced?

dglazer:
It looks like there are two copies of a couple of new image files -- can we delete whichever copy isn't being used?

docs/asciidoc/prefix_register_1.png 
docs/asciidoc/prefix_register_2.png 
docs/prefix_register_1.png 
docs/prefix_register_2.png

docs/asciidoc/back_matter.adoc
```diff
@@ -39,3 +37,94 @@ This spec defines a standard **Data Repository Service (DRS) API** (“the yello
...
+
+=== Security with Compact Identifiers
+
+As mentioned earlier, identifiers.org/n2t.net performs some basic verification of new prefixes and provider code mirror registrations on their sites.  However, system builders may have certain security compliance requirements and regulations that prohibit relying on an external site for resolving compact identifiers.  In this case, systems under these security and compliance constraints may wish to whitelist certain compact identifier resolvers and/or vet records from identifiers.org/n2t.net before enabling in their system.
```
dglazer, line 112: I like the caveat here about giving up control over security, but I don't understand what guidance we're providing to whom. For server builders, they should either decide they're comfortable with CI-based URIs or not -- that's a one-time decision, and doesn't give them any control over whitelisting or mirroring. For client builders, if they're not allowed to access external sites, I think that means they can't handle CI-based URIs (which technically means they're no longer DRS-compliant) -- what are we suggesting they whitelist, and how would that help?

docs/asciidoc/back_matter.adoc
```diff
@@ -39,3 +37,94 @@ This spec defines a standard **Data Repository Service (DRS) API** (“the yello
...
+
+=== Accession Encoding to Valid DRS IDs
+
+The compact identifier format used by identifiers.org/n2t.net does not percent-encode reserved URI characters but, instead, relies on the first ":" character to separate prefix from accession. Since these accessions can contain any characters, and characters like "/" will interfere with DRS API calls, you _must_ percent encode the accessions extracted from DRS compact identifier-based URIs when using as DRS IDs in subsequent DRS GET requests.  An easy way for a DRS client to handle this is to get the initial DRS object JSON response from whatever redirects the compact identifier resolves to, then look for the `self_uri` in the JSON, which will give you the correctly percent-encoded DRS ID for subsequent DRS API calls such as the `access` method.
```
dglazer, line 116: I'm confused by this guidance -- given that DRS URIs are already required to be URI-friendly, there shouldn't be anything left to encode. Unless I'm missing something, let's delete this section.

TimothyTickle:
## Driver Vote

* Driver Project: HCA DCP
* Vote: +1
* Who: Timothy Tickle
* Comments: No additional comments

ianfore:
----
## Driver Vote

* Driver Project: NCI Cancer Research Data Commons
* Vote: +1
* Who: Ian Fore. [Data Commons Framework](https://dcf.gen3.org) and [NCI Cloud Resources](https://datascience.cancer.gov/data-commons/cloud-resources) were canvassed.
* Comments: A good approach worked out through a good consultative process. The question remains how mutable/logical-level will be identified/referred to, but that's a separate problem as immutability is a central principle in DRS.
----

mattions:
### Implementer vote
- Vote: +1
- Implementer: Seven Bridges (developing BioDataCatalyst by SB (TOPMed), CGC (NCI Cloud Resources) and Cavatica (Kids First)

briandoconnor:
## Driver Vote

* Driver Project: TOPMed
* Vote: +1
* Who: Brian O'Connor, Albert Smith
* Comments: This PR will allow us to interoperate with Gen3 using the GA4GH DRS API while still supporting the compact identifiers the BD Cat platform (and others) want to support.  Please release this soon, at least four NIH ICs are waiting on this in order to interoperate with each other.

docs/asciidoc/front_matter.adoc
briandoconnor: OK, removed

docs/asciidoc/front_matter.adoc
briandoconnor: Doesn't seem like this is an issue that's worth delaying 1.1 and it's not central to any functionality in the proposal so I'm just removing it.  I haven't seen any driver project care either way.

docs/asciidoc/front_matter.adoc
briandoconnor: Yep, done!   I captured the first todo (self URL reference) in a ticket.

docs/asciidoc/front_matter.adoc
briandoconnor: Done

docs/asciidoc/back_matter.adoc
briandoconnor: Done, I indented this and the previous section

docs/asciidoc/back_matter.adoc
briandoconnor: Done

docs/asciidoc/back_matter.adoc
briandoconnor: I think it's worth pointing out in the main spec but I agree that it adds clutter and, more importantly, something that may age very poorly.  So I left in the mention of this via the first paragraph but then put the actual example in the background doc as you suggested.

docs/asciidoc/back_matter.adoc
briandoconnor: Ian brought up the importance of the last line here "Alternatively, system builders may decide to directly mirror the registries themselves, instructions are provided on the identifiers.org/n2t.net websites."

I went ahead and streamlined this section to describe caching a bit more but without just restating the line in the main text doc

docs/asciidoc/front_matter.adoc
briandoconnor: done

docs/sample_jsons/drsobject.24207d90-c5b8-4fac-8c07-6426521dfeac.json
briandoconnor: removed

briandoconnor:
> It looks like there are two copies of a couple of new image files -- can we delete whichever copy isn't being used?
> 
> docs/asciidoc/prefix_register_1.png
> docs/asciidoc/prefix_register_2.png
> docs/prefix_register_1.png
> docs/prefix_register_2.png

This is a build system strangeness... It didn't seem to work properly when I had the images in one location and not both. I'm not going to worry about the duplication for now since the build system will be totally changed in the near future.

docs/asciidoc/back_matter.adoc
briandoconnor: I rephrased this to be "builders of systems that consume and resolve DRS URIs".  Specifically, I added this language in following a conversation with David Bernick at the Broad about the ability of FedRAMP/FISMA Moderate environments that will wish to support DRS 1.1 URI resolution within their security boundary.  By adding this language I'm hoping to give systems like these some guidance on maintaining their compliance while still supporting DRS 1.1

docs/asciidoc/back_matter.adoc
briandoconnor: Identifiers.org/n2t.net accessions do not use encoding while DRS IDs (and any calls made to DRS API endpoints) do.  Hence this description.  We can remove it but it will leave some key behavior underspecified and will likely cause confusion by implementers.  I strongly recommend leaving this and discussing with our drivers/implementers to clarify language.

soumyadipDe:
## Implementer vote

- Vote: +1
- Implementer: Soumyadip De, Andrea Cristofori, Tony Wildish (RDSDS by EMBL-EBI)

MichaelLukowski:
## Driver Vote

* Driver Project: NCI GDC
* Vote: +1
* Who: Michael Lukowski, Robert Grossman, Jiaqi Liu

shaikh-rashid:
# Driver Vote
- Driver Project: CanDIG
- Vote: 0
- Who: Shaikh Rashid
- Comment: Overall we are neutral to the change in URI scheme

juhtornr:
### Driver Vote

- Driver Project: ELIXIR Beacon
- Vote: 0
- Who: Juha Törnroos
- Comment: Inside ELIXIR we have been planning to use checksums as identifiers but to my understanding this is not (completely) incompatible with that so we are neutral on this.

kwrodarmer:
NCBI/NLM/NIH implementation endorses the PR.

Comments - our server is unlikely to issue ids in compact format.

briandoconnor:
Given the votes, I'm merging this into develop.

Issue #348, opened 2021-04-12
Support POST'ing of Passports to DRS endpoints


briandoconnor:
**DEADLINE**: please comment on the PR #363 before the next Cloud call on Sept 13th.  We will merge this PR then if there are no blocker comments.

In the TOPMed (BioData Catalyst) driver project (and applicable to GDC and NCI Drivers) we have been exploring different ways of interacting between Passport systems ([NIH RAS](https://auth.nih.gov/docs/RAS/serviceofferings.html) in this case) and a DRS server.  There are multiple proposals for how a Passport with Embedded Visas should be handed to a DRS server and BDCat has documented 3 different approaches we're considering [here](https://docs.google.com/document/d/1T3uYGS2yZflDbLRbG4uxhi8ICqk9C9xWPmJ0DQpFvDU/edit#heading=h.5atd0vqkj5vq).  We've largely settled on an approach where RAS Passports will be POST'd directly to a DRS server, see [here](https://docs.google.com/document/d/1amOGLwAbKkMSU6up_dHGhsuUF9SVFlllO0qjQSJYuVI/edit#heading=h.4uzh9l6x5chi).

In the RAS passport to DRS server flow (option 2, scenario 1) we would need a way to POST a RAS Passport to a DRS server and right now the /objects/{object_id}/access/{access_id} only supports GET.

This issue is accompanied by a feature branch [feature/issue-348-post-passport-v2](https://github.com/ga4gh/data-repository-service-schemas/tree/feature/issue-348-post-passport-v2) where I'm trying to update the schema to accommodate posts.

You can follow builds on [Travis](https://app.travis-ci.com/github/ga4gh/data-repository-service-schemas/builds)

Built docs can be found at [preview/feature/issue-348-post-passport-v2/docs](https://ga4gh.github.io/data-repository-service-schemas/preview/feature/issue-348-post-passport-v2/docs)

Swagger schema can be downloaded from the top of the docs above and visualized at https://editor.swagger.io

The PR: https://github.com/ga4gh/data-repository-service-schemas/pull/363

ianfore:
Would this just be for Passport? Or are we offering POST for the other auth mechanisms that currently exist e.g. the payloads that are currently passed via Bearer tokens?

briandoconnor:
PR was merged so closing

Issue #230, opened 2019-02-06
Update to OpenAPI 3.0


sarpera:
[Current DRS schema definition](https://github.com/ga4gh/data-repository-service-schemas/blob/develop/openapi/data_repository_service.swagger.yaml) is based on OpenAPI (fka swagger) 2.0. Shall we update it to conform to [OpenAPI 3.0](https://swagger.io/blog/news/whats-new-in-openapi-3-0/) (the latest version of the OpenAPI Specification)?

Immediate benefit of the update is to have [new data model capabilities](https://swagger.io/docs/specification/data-models/) which would make the ongoing efforts on creating PRs a bliss.

ddietterich:
I agree. It would be good to use 3.0.

The only downside I know is that swagger-codegen does not yet support 3.0. Anyone planning to use that tool?

denis-yuen:
We do, but we've had successful experiments with swagger-codegen 3.0.4 https://github.com/swagger-api/swagger-codegen/releases/tag/v3.0.4 which does support OpenAPi 3.0.

geoffjentry:
To echo - the conventional wisdom in the cloud workstream has been what @ddietterich stated, but is obviously going to be a moving target. My $0.02 is that as long as the infrastructural challenges are surmountable that we should go the 3.0 route

briandoconnor:
What impact will this have on build?

jaeddy:
Just noticed that **swagger2markup** (which we use to autogen documentation during the build process) doesn't currently support OpenAPI 3 — there might be some potential workarounds we could try:
https://github.com/Swagger2Markup/swagger2markup/issues/340

I think we'd need to adjust the Swagger UI HTML template as well, but that should be a quick fix.

natanlao:
If it would ease the transition, OpenAPI 3.0 and SmartAPI schema definitions can be generated from the Swagger schema by running `make schemas` in the project root, which will create `openapi/data_repository_service.openapi.yaml` and `openapi/data_repository_service.smartapi.yaml`.

denis-yuen:
FYI @garyluu

sarpera:
I’ve tried a few things and have some question:

First of all, seems like the instructions are out of date here: https://github.com/ga4gh/data-repository-service-schemas/blob/develop/README.md or
https://ga4gh.github.io/data-repository-service-schemas/#getting-started

[...]

PR #349, opened 2021-04-12
Feature/issue 348 post passport

briandoconnor:
See Issue #348

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ dist: trusty
 language: python
 python:
 - '3.6'
-- '2.7'
+#- '2.7'
 
 # Specifying `branches.only = ['master']` can cause tagged builds to
 # not deploy. See travis-ci/travis-ci#2498 and travis-ci/travis-ci#1675.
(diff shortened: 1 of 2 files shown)


dglazer:
OpenAPI question -- do we need to duplicate the content of the POST and GET endpoints? It would be a lot easier to read (and to maintain) if we could share the duplicated part.

briandoconnor:
Bigger issue @dglazer is this is swagger 2.0... not OpenAPI which is what we should move towards as a project.  I'd prefer to not invest a lot of time figuring out how to do this in the old swagger language and focus on cleaning it up in the move to OpenAPI 3

dglazer:
I agree we should move to OpenAPI 3 asap -- @jb-adams to make sure that's on the list.
I'm not sure that will help with this particular issue, but I agree it's not worth debugging in swagger syntax.

jb-adams:
Looks like #230 aims to migrate to OpenAPI 3. @jaeddy @uniqueg has there been any movement on this lately? I can take this branch if you have no bandwidth currently

briandoconnor:
@jb-adams , it would be fantastic to move the OpenAPI 3 work forward and coordinate that with @jaeddy as part of the trans GA4GH effort to have common doc build approaches across our repos.

uniqueg:
Thanks @jb-adams that would be appreciated. Very little bandwidth here in May and June, unfortunately.

briandoconnor:
I'm closing this PR in favor of #363 which uses OpenAPI 3

PR #362, opened 2021-08-03
Implement /service-info endpoint in DRS

jb-adams:
[Documentation Preview](https://ga4gh.github.io/data-repository-service-schemas/preview/feature/issue-309-add-service-info/docs/)

This PR aims to add the GA4GH standardized `/service-info` endpoint to DRS.

Some questions:
- Should we extend the base Service Info payload format with DRS-specific attributes to further inform clients of DRS server properties? Ideas include:
  * Service location/region
  * Supported access methods / clouds that referenced DrsObjects can be accessed on (e.g. does this DRS service point me to AWS S3 objects? Google Cloud storage? Both? Others?)
  * URLs to passport broker services yielding visas to controlled datasets 
  * additional metadata keys returned from the `/objects/{id}` endpoint and what they mean
- At the end of the spec there is a section called "Future DRS Versions and Service Registry/Info." Should this section be removed/changed?
- Should we say anything about service registry? E.g. what is the base URL one should register in a service registry for a given DRS service?
- I have added an additional constraint to the `type` property, stating that the `artifact` must display `drs` (rather than `beacon`, `htsget`, etc.) Should we also add constraints for what `versions` of the specification services are allowed to display? E.g. any current or historical release of the DRS specification: `"version": "1.0.0"`, `"version": "1.1.0"`, `"version": "1.2.0"` etc


briandoconnor:
@jb-adams can you add a link to the generated docs?

jb-adams:
From yesterday's Cloud WS meeting, we agreed to:
* Only include the base service-info response for now, and not extend the response payload with DRS-specific attributes (these can be added in a later PR)
* Explicitly mention how the `URL` of a DRS service must be registered in a service registry. DRS services must be registered according to a consistent pattern in order for federation to work properly. I propose that the registered `URL` should be up to but not including the standardized `/ga4gh/drs/v1`. Since the registry entry will indicate it's a `drs` service, a client can easily reconstruct `{URL}/ga4gh/drs/v1/objects/{OBJECT_ID}` to get an object, or `{URL}/ga4gh/drs/v1/service-info` to get service info

openapi/components/responses/200ServiceInfo.yaml
```diff
@@ -0,0 +1,7 @@
+description: Retrieve info about the DRS service
+content:
+  application/json:
+    schema:
+      allOf:
+        - '$ref': https://raw.githubusercontent.com/ga4gh-discovery/ga4gh-service-info/v1.0.0/service-info.yaml#/components/schemas/Service
+        - '$ref': '../schemas/DrsServiceInfo.yaml'
```
tschaffter, line 7: What about naming the schema "DrsService"? The "Info' part is not very informative. For example, a user schema would typically be called "User", not "UserInfo".

jb-adams: Good point @tschaffter , and also because the base schema is `Service` and not `ServiceInfo`. I'll update

PR #363, opened 2021-08-08
Feature/issue 348 post passport v2

briandoconnor:
**DEADLINE**: please comment on the PR #363 before the next Cloud call on Sept 13th.  We will merge this PR then if there are no blocker comments.

See Issue #348 for the background on this PR

Note, I'm closing PR #349 and replacing it with this one to move to OpenAPI 3.

Some questions for the group:

1. Is the PassportAuth SecuritySchemes done correctly?  See data_repository_service.openapi.yaml and this documentation from OpenAPI: https://swagger.io/docs/specification/authentication/ 
1.  Is there a way to not have duplication information between the GET and POST versions of the same endpoint?  I tried multiple approaches to this and none worked.  I need someone better at OpenAPI 3 to help with this.

### DIFF
--- a/openapi/paths/objects@{object_id}.yaml
+++ b/openapi/paths/objects@{object_id}.yaml
@@ -62,7 +62,7 @@ post:
         schema:
           type: object
           properties:
-            tokens:
+            passports:
               type: array
               items:
                 type: string
--- a/openapi/paths/objects@{object_id}.yaml
+++ b/openapi/paths/objects@{object_id}.yaml
@@ -32,6 +32,8 @@ post:
 
     Method is a POST to accomodate a JWT GA4GH Passport sent in the formData in order to authorize access.
   operationId: PostObject
+  security:
+    - PassportAuth: []
   responses:
     200:
       $ref: '../components/responses/200OkAccess.yaml'
--- a/openapi/paths/objects@{object_id}.yaml
+++ b/openapi/paths/objects@{object_id}.yaml
@@ -24,3 +24,44 @@ get:
   tags:
     - Objects
   x-swagger-router-controller: ga4gh.drs.server
+
+post:
+  summary: Get info about a DrsObject through POST'ing a Passport.
+  description: >-
+    Returns object metadata, and a list of access methods that can be used to fetch object bytes.
+
+    Method is a POST to accomodate a JWT GA4GH Passport sent in the formData in order to authorize access.
+  operationId: PostObject
+  responses:
+    200:
+      $ref: '../components/responses/200OkAccess.yaml'
+    202:
+      $ref: '../components/responses/202Accepted.yaml'
+    400:
+      $ref: '../components/responses/400BadRequest.yaml'
+    401:
+      $ref: '../components/responses/401Unauthorized.yaml'
+    403:
+      $ref: '../components/responses/403Forbidden.yaml'
+    404:
+      $ref: '../components/responses/404NotFoundAccess.yaml'
+    500:
+      $ref: '../components/responses/500InternalServerError.yaml'
+  tags:
+    - Objects
+  x-swagger-router-controller: ga4gh.drs.server
+  parameters:
+    - $ref: '../components/parameters/ObjectId.yaml'
+    - $ref: '../components/parameters/Expand.yaml'
+  requestBody:
+    required: true
+    content:
+      multipart/form-data:
+        schema:
+          type: object
+          properties:
+            tokens[]:
+              type: array
+              items:
+                type: string
+              description: the encoded JWT GA4GH Passport that contains embedded Visas.  The overall JWT is signed as are the individual Passport Visas.
--- a/openapi/paths/objects@{object_id}.yaml
+++ b/openapi/paths/objects@{object_id}.yaml
@@ -56,14 +56,4 @@ post:
     - $ref: '../components/parameters/ObjectId.yaml'
     - $ref: '../components/parameters/Expand.yaml'
   requestBody:
-    required: true
-    content:
-      application/json:
-        schema:
-          type: object
-          properties:
-            passports:
-              type: array
-              items:
-                type: string
-              description: the encoded JWT GA4GH Passport that contains embedded Visas.  The overall JWT is signed as are the individual Passport Visas.
+    $ref: '../components/requestBodies/Passports.yaml'
--- a/openapi/paths/objects@{object_id}.yaml
+++ b/openapi/paths/objects@{object_id}.yaml
@@ -58,11 +58,11 @@ post:
   requestBody:
     required: true
     content:
-      multipart/form-data:
+      application/json:
         schema:
           type: object
           properties:
-            tokens[]:
+            tokens:
               type: array
               items:
                 type: string
--- a/openapi/paths/objects@{object_id}@access@{access_id}.yaml
+++ b/openapi/paths/objects@{object_id}@access@{access_id}.yaml
@@ -4,30 +4,27 @@ description: >-
 
   This method only needs to be called when using an `AccessMethod` that contains an `access_id`
   (e.g., for servers that use signed URLs for fetching object bytes).
-operationId: GetAccessURL
-responses:
-  200:
-    $ref: '../components/responses/200OkAccess.yaml'
-  202:
-    $ref: '../components/responses/202Accepted.yaml'
-  400:
-    $ref: '../components/responses/400BadRequest.yaml'
-  401:
-    $ref: '../components/responses/401Unauthorized.yaml'
-  403:
-    $ref: '../components/responses/403Forbidden.yaml'
-  404:
-    $ref: '../components/responses/404NotFoundAccess.yaml'
-  500:
-    $ref: '../components/responses/500InternalServerError.yaml'
-tags:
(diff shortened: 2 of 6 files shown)


openapi/paths/objects@{object_id}.yaml
```diff
@@ -23,4 +23,47 @@ get:
...
+  tags:
+    - Objects
+  x-swagger-router-controller: ga4gh.drs.server
+  parameters:
+    - $ref: '../components/parameters/ObjectId.yaml'
+    - $ref: '../components/parameters/Expand.yaml'
+  requestBody:
+    required: true
+    content:
+      multipart/form-data:
```
jb-adams, line 61: suggest changing this to `application/json` if possible, since then it will be easier to add in `selection`, etc. as another attribute

openapi/paths/objects@{object_id}.yaml
```diff
@@ -23,4 +23,47 @@ get:
...
+        schema:
+          type: object
+          properties:
+            tokens[]:
```
jb-adams, line 65: This should just be `tokens` and doesn't need the `[]`. It's already clear from `type: array`.

jb-adams, line 65: Is `tokens` the best attribute name to use here? Since we are trying to think long term interop with other standards, we don't want a collision if another API is using `tokens` to mean something else.

Would `passport_tokens`, `passportTokens` be more appropriate?

openapi/paths/objects@{object_id}.yaml
```diff
@@ -23,4 +23,47 @@ get:
...
+    404:
+      $ref: '../components/responses/404NotFoundAccess.yaml'
+    500:
+      $ref: '../components/responses/500InternalServerError.yaml'
+  tags:
+    - Objects
+  x-swagger-router-controller: ga4gh.drs.server
+  parameters:
+    - $ref: '../components/parameters/ObjectId.yaml'
+    - $ref: '../components/parameters/Expand.yaml'
```
jb-adams, line 57: I think having query string parameters in a POST request might violate some web conventions and/or standards. RESTful design generally favours using query string for GET requests, and request body (without any query string parameters) for POST requests (e.g. HTML forms with an action of `POST` take all parameters and submits them in the request body).

Perhaps we can have `expand` as a valid request body param rather than query string param, e.g.
```
{
  "expand": true,
  "tokens": [
    "...",
    "..."
  ]
}
```

openapi/paths/objects@{object_id}@access@{access_id}.yaml
```diff
@@ -26,4 +56,19 @@ get:
...
-  x-swagger-router-controller: ga4gh.drs.server
\ No newline at end of file
+  x-swagger-router-controller: ga4gh.drs.server
+  parameters:
+    - $ref: '../components/parameters/ObjectId.yaml'
+    - $ref: '../components/parameters/AccessId.yaml'
+  requestBody:
+    required: true
+    content:
+      multipart/form-data:
```
jb-adams, line 66: See comment about using `application/json`

openapi/paths/objects@{object_id}@access@{access_id}.yaml
```diff
@@ -26,4 +56,19 @@ get:
...
+        schema:
+          type: object
+          properties:
+            tokens[]:
```
jb-adams, line 70: See comment about using `tokens` instead of `tokens[]`

jb-adams (commented):
Looks good so far, my requested changes generally revolve around formatting the `POST` request body as `application/json` rather than `multipart/form-data`

BinamB:
I am not completely convinced that DRS should accept multiple passports.  This deviates from the original idea of "One token to rule them all".  I can see how this might be useful when someone might not want to repackage their passports but I think accepting multiple passports defeats the purpose of a passport. This wouldn't be any better than doing traditional OIDC with extra steps (i.e. just sending in a list of access tokens from different systems to a DRS server). The purpose of a passport is to "transport a researcher's digital identity and permissions across organizations, tools, and environments". A single passport is the best solution, it doesn't work if we're keeping track of multiple passports. 

As an implementer, this is also concerning since this does not look like it would scale very well. I think there are already scalability concerns that we're trying to solve with a single passport. Passports do tend to get very large and sending multiple large passports to a DRS server might result in a poor user experience. I think scalability should be taken into consideration so that we don't have to come back to square one after there have been a couple of implementations.

openapi/paths/objects@{object_id}.yaml
briandoconnor: sounds good, making that change

openapi/paths/objects@{object_id}.yaml
briandoconnor: sounds good, making that change

briandoconnor:
@BinamB, I agree with your general points.  We talked during the GA4GH Cloud Meeting (can't remember the date) on your feedback above.  Just putting some notes in here to record the info on the PR:

* I and others agree about wanting to support repackaged visas... but we have to listen to drivers that asked for the array of passports.  This gives us flexibility to, say, send a RAS passport unmodified to a DRS server along with a Gen3/Terra/SBG/other passport that's repackaged many different visas.  So we can support 1) multi-passports, 2) a single repackaged passport, 3) or a mix of the two
* in the longer term, downscoping passports will ultimately solve this problem better... but we're not there yet.  So this approach allows us to move forward with NIH projects using RAS with relatively little impact on the DRS spec.  But longer term, downscoping passports should really be the ultimate solution for your performance concerns (e.g. just send the only visa needed to access a given DRS object)
* having this be an application/json means we can add information to the POST in the future, the selection object being worked on by Kurt for example, that should allow us to extend this to support passport downscoping in the future

openapi/paths/objects@{object_id}@access@{access_id}.yaml
briandoconnor: done

openapi/paths/objects@{object_id}@access@{access_id}.yaml
briandoconnor: done

openapi/paths/objects@{object_id}.yaml
briandoconnor: I'm on the fence.  I think "yes" but then I wonder, what if we want to support alternative token JWT formats?  Beyond just passport 1.0.  Calling it tokens is more generic.  But if everyone wants "passports" then I'm fine with that.

openapi/paths/objects@{object_id}.yaml
briandoconnor: Want to talk about this on today's call?

openapi/paths/objects@{object_id}.yaml
jb-adams: `tokens` renamed to `passports` based on discussion on Aug 23rd Cloud call

openapi/paths/objects@{object_id}.yaml
jb-adams: @briandoconnor forgot about this one at yesterday's call, only remembered now unfortunately. Is there any way we can solicit feedback on this item?

jb-adams:
rebased this branch onto `develop`, bringing in the recently closed Service Info PR

kwrodarmer (changes requested):
One of the omissions from passport v1 design that has a negative impact is the lack of labels on tokens. Passports group tokens as an array - which works, but in the case of multiple tokens it literally requires cryptographically testing the signature of each object in order to figure out which is which.

A natural type of label would be a type, or maybe an issuer. A less intuitive but very useful label is the key id, assuming that it is globally scoped. Regardless, if the bundle of tokens posted to DRS could be given a label rather than just put into an array, that would be very useful.

BinamB:
@kwrodarmer Im assuming this would look something like a key:value pair where the key is the issuer and value is the token. I see how this increases performance but I don't see how we can implement it in the DRS server. Correct me if I'm wrong but this would mean that each DRS object would have to map to an issuer to decide on which token to use?

kwrodarmer:
Using something like you describe:
`[ "name1:jwt1", "name2:jwt2" ]`
would be a way to do it. Also, using a JSON object as
`{ "name1" : "jwt1", "name2" : "jwt2" }`
would work.

> each DRS object would have to map to an issuer to decide on which token to use?  

I'm not sure I understand. Are you saying they don't?

kwrodarmer:
As a follow on, there is an issue around labels on tokens - where do they come from, who assigns them, can they be reassigned, etc. This may prove to complicate matters too much to incorporate them into the current PR and DRS version. We have discussed it but not addressed it within the passport specification, so maybe we should revisit the issue in a wider scope. And @BinamB , maybe that is what you were saying before.

BinamB:
Yeah, that is kinda what I was trying to say. It also complicates things when the system itself is trying to give access through their own Passports for something like developer access. But I think we're heading in a good direction by using a JSON instead of a list. Maybe we can save this for passports v2 but something I've been thinking about is instead of how we have a list of permissions right now in v1 we could have a JSON/Dictionary of it which would be a lot faster/easier for a DRS server to search through.

Issue #334, opened 2020-10-20
DRS bulk requests


briandoconnor:
Some DRS implementers/users have requested the ability to make DRS requests in bulk for multiple DRS URIs.

See NHGRI AnVIL and the use of Galaxy in that project.  Hope they can fill in more information about the use case here.

luke-c-sargent:
Late to the party, but to fill in some Galaxy-centric details: there are AnVIL workspaces that contain tables with thousands of DRS URIs; sometimes there exists associated metadata within these tables Galaxy can use to pre-populate fields and delay resolution til the point of actual file acquisiton, sometimes there is not. To show the user the nature of the data they are browsing in the latter case (and not just a big guid), that will require thousands of individual HTTP connections to the resolver. It seems like batch resolution is a win for everyone re: infrastructure strain and user experience in any circumstance where there is more than one DRS URI to resolve.

Bulk resolution could as simple as a body containing a list of DRS URIs. If there is some concern about this being non conformant to the DRS spec (I vaguely recall this being mentioned when I brought it up in a call), perhaps implementing something like [Google storage batching](https://cloud.google.com/storage/docs/json_api/v1/how-tos/batch) via [multipart message types](https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html) would allow the underlying system to operate on in individual basis while still providing the time / bandwidth savings. This might also be useful in situations where a system wants to aggregate requests from multiple users, with each part having its own auth token (Example: AnVIL users can click a button to resolve DRS URIs to see its metadata; if there were tens of thousands of concurrent users making these requests, the underlying systems could bundle all of their requests into multipart packets with individual auth tokens, sort the results, and shave off precious UI lag seconds).

These are just some thoughts; of course the best folks to formulate the correct approach are you all! Thanks for keeping this in your dev thought processes, I'm keen to see how this progresses and happy to answer any questions re: the Galaxy on AnVIL side of things.

jb-adams:
[...]

PR #365, opened 2021-08-16
Feature/issue 334 drs bulk requests

briandoconnor:
### Goal

We want to have this merged into develop for DRS 1.4 for the 2023 Plenary

### Background 

See Issue #334 

A PR request for bulk operations in DRS.  As of 5/22/23 we have the complete set of bulk endpoints for authorization information, DRS IDs, and DRS access methods.  This PR does not include pagination nor does it include explicit pairing of passports to the output of a bulk response.  For pagination, I think we should move bulk forward and cleanly separate out pagination as a different feature/PR.  For pairing, the feature branch now has bulk options support, so you can 1) ask the DRS server what authorization mechanisms are needed for a bulk list of DRS IDs and then 2) make one or more bulk requests where you pair the correct passport(s) or bearer tokens in your request.  

### Built Docs

See the built doc here: https://ga4gh.github.io/data-repository-service-schemas/preview/feature/issue-334-drs-bulk-requests/docs/#tag/Objects/operation/GetBulkObjects/ 

### For more information...

See the [Hackathon notes](https://docs.google.com/document/d/1h1EP8Z8SyTS_T4pwMVO_SfvEynHw2U7hNk8I8qhp4OA/edit#) from our June FASP hackathon.  Including this [gist](https://gist.github.com/mbarkley/6eec9ee9bd77ffc6a022b0459c0fb255)

See also[ Feature/issue 334 bulk requests nobundle](https://github.com/ga4gh/data-repository-service-schemas/pull/377) from Ian and [DRS bundle contents pagination](https://github.com/ga4gh/data-repository-service-schemas/pull/367) from Jeremy


openapi/data_repository_service.openapi.yaml
```diff
@@ -103,8 +103,12 @@ x-tagGroups:
 paths:
   /objects/{object_id}:
     $ref: ./paths/objects@{object_id}.yaml
+  /bulk/objects:
```
jb-adams, line 106: How about making this endpoint at just `POST /objects` rather than `POST /bulk/objects` ?

openapi/paths/bulkobjects@access@{access_id}.yaml
```diff
@@ -0,0 +1,40 @@
...
+    - Objects
+  x-swagger-router-controller: ga4gh.drs.server
+  requestBody:
+    required: true
+    content:
+      application/json:
+        schema:
+          type: object
+          properties:
+            object_ids:
```
jb-adams, line 36: How about renaming this property to `selection` to harmonize with the DRS + Passport downscoping thread. 

@kwrodarmer @mbarkley WDYT?

openapi/components/responses/200OkDrsObjects.yaml
```diff
@@ -0,0 +1,10 @@
+description: The `DrsObjects` were found successfully
+content:
+  application/json:
+    schema:
+      type: object
+      properties:
+        drsobjects:
```
jb-adams, line 7: Some thoughts on the response payload of a bulk DRS request:
1. Currently `drsobjects` is a flat list, would it be advantageous to make it an object with the supplied IDs as keys, and the corresponding DRS Object as the value?
2. There could be a mix of valid and invalid IDs in the overall selection (e.g. no resource by an ID, client not authorized to access that DRS Object). Should we make a slightly more complex object that clearly outlines which IDs could not be returned as Objects and why?

Taking the above points together, here is a proposed example response payload:
```
{
  "summary": {
    "requested": 5,
    "loaded": 2,
    "unloaded": 3
  },
  "loadedDrsObjects": {
    "123": {
      "id": "123",
      "name": "DRS Object 123",
      ...
    },
    "456": {
      "id": "456",
      "name": "DRS Object 456",
      ...
    }
  },
  "unloadedDrsObjects": {
    "777": 404,
    "778": 404,
    "779": 401 
  }
}
```

In the above example, the response payload summarizes how many IDs were passed, how many DRS Objects were successfully retrieved, and how many were unsuccessful. Each retrieved DRS Object can be referenced by its ID. Each non-retrieved DRS Object comes with an explanation (in the form of HTTP code) why it wasn't retrieved.

It's a more complex payload overall, but provides more information to the client about how the request was processed.

openapi/data_repository_service.openapi.yaml
briandoconnor: OK, I'll make that change.

We do need to figure out how to enable multi-part POST here so it can be used with Passports.  this is a TODO

openapi/paths/bulkobjects@access@{access_id}.yaml
briandoconnor: So "object_ids" --> "selection"?  I'm fine with that...

openapi/components/responses/200OkDrsObjects.yaml
briandoconnor: Let's discuss on today's call

patmagee: A few comments on the proposed format:

## Summary
Depending on the number of items that are requested and where we end up falling on pagination in a bulk request, the summary information may not be known ahead of time when the initial response is sent. Its possible this number may be per page, instead of in total.


## Results Format
It may be nice to return the objects as an array with the ordering mirroring the order of the requested Id's This leads to less surprises and allows the client to iterate over the objects in the order they expect.  Additionally, since this is a bulk operation we could interleave the errors that are encountered directly in the array at the appropriate index. (this is using the error message format from beacon [here](https://github.com/ga4gh-beacon/specification/blob/master/beacon.yaml#L619)

```json
{
  "drsObjects" [
    {
      "id": 456,
      "name":  "..."
    },
    {
      "id": 456,
      "error":  {
        "errorCode": 404,
        "errorMessage": "DRS object does not exist"
      }
    }
  ]
}
```

openapi/paths/bulkobjects@access@{access_id}.yaml
ianfore: It started as selection and, with some discussion,  we consciously went to object_ids as there may be other use cases than "selection" for how you get to a list of ids that you want to request. object_ids is neutral as to use case, rather than projecting a particular use case on to DRS which has some generality.

ianfore:
See also PR#377 - https://github.com/ga4gh/data-repository-service-schemas/pull/377

briandoconnor:
Thanks @bcli4d this is great feedback.  I'll make these changes/clarifications since they all sound totally reasonable.

briandoconnor:
How will this work with the opt PR that looks at per object authZ/passports?  Will need to revise this.

briandoconnor:
To keep this on track for Connect in April 2023:

Champion → @briandoconnor  (but help from @ianfore and others to merge in other PR ideas would be great, e.g. #377 and #367 
Deadline for revisions: I want to have any changes landed by 2/13 so folks can do a final round of comment
Deadline for comments: 2/27 for folks to weigh in on what they see in this PR
Merge date: 3/13 at latest to merge into develop so this can be part of DRS 1.4 release branch

briandoconnor:
FYI, this branch is building properly even though it's a PR.  So I'm not going to use (and will delete) the v2 of this feature branch.

briandoconnor:
Added 413 error based on @bcli4d 's comment above

briandoconnor:
So looping back on the remaining issues/comments here:

- bulk_objects_ids and bulk_object_access_ids array length max needed in /service-info, should be mandatory and describe the max length the bulk requests can be for this DRS server --> added maxBulkRequestLength to the /service-info endpoint 
- "if some but not all objects were resolved" --> the response includes "unresolved_drs_objects" to show which are not resolved.  If none are resolved then the server can return 404
- better descriptions instead of "Array of Strings" --> I added description text

briandoconnor:
@dglazer gave the feedback of what happens when the DRS server doesn't support bulk?  Is the field still required?  To be "1"?

bcli4d:
I think defaulting to 1 in the absence of the field is OK.

dglazer:
I'm "+0" on my own suggestion -- I think it's clean to say "you don't have to support batch, and if you don't you shouldn't have to specify a max batch size".

But I'm only +0 since, as Brian pointed out in person, my initial concern about breaking backwards compatibility was overblown, since compatibility only kicks in after you've already decided to upgrade to a new API version.

briandoconnor:
Just as a heads up, I'm planning on merging this into develop and opening a 1.4 release branch (which will have a request for comment period by drivers/implementers) by the end of GA4GH Connect 2023.  Unless I hear an urgent problem that needs to be fixed of course...

briandoconnor:
Thanks everyone, I don't see any remaining issues.  Merging into develop now.  If you see more issues I'll open a release branch for comments.  Always remember, this is going to develop now so if you see an issue you can always open up a new PR to fix it before we finalize 1.4

Issue #366, opened 2021-09-14
contents array pagination of DRS Object bundle


jb-adams:
Related to #342 (scaling concerns) , specifically #325 (pagination in DRS)

There may be multiple types of pagination that the DRS spec eventually incorporates. 2 foreseeable instances include:
1. Pagination through the `contents` of a single DRS Object bundle
2. Pagination through the result of a large selection/multi-id/batch request

This issue focuses on (1). The current `/objects/{object_id}` can be updated to allow for `contents` array paging

Issue #325, opened 2020-07-13
DRS paging


briandoconnor:
Ideas:
Can we include...
has_more
next (URL to the next page)
previous (URL to the previous page)
page_count: number of pages (can we make this optional?)
items_per_page: number of bundle items per page
These would be present in every response.
Also consider adding a requested page size

dG: see also https://cloud.google.com/apis/design/design_patterns#list_pagination for a slightly simpler pagination API style guide.

Next steps
Common approach, see link above
check with TASC Force -- Is GA4GH doing this in a consistent way?  Is there a common pattern we should use across all our APIs?
Who will implement this?  U. Chicago?  Other groups?  We need at least one driver.
U. Chicago implementer 
CRDC, GDC/IDC
TOPMed
EMBL imaging site?

kaushik-work:
A provision for an index is something that would be generally useful. A somewhat flexible structure that mapped a key to a position in the bundle. I imagine for large bundles it is inefficient to jump from page to page looking for something.

pgrosu:
Besides the offset-with-limit approach Kaushik mentioned, you probably want a sort feature so you can perform a seek-filter (i.e. retrieve after some id) with a limit.

ianfore:
Discussed within CRDC Imaging and Data Commons framework. Determined that 'standard' pagination capability is required.
'Standard' pagination capability would amount to what was suggested at the top of the issue. Copied and annotated below. 
However, rather than specifying all those from scratch suggest that:
- There are a number of standard approaches already in use, both in and out of GA4GH. Suggest that adopting one of these rather than coming up with a new one 
- A common approach to pagination would be desirable across GA4GH APIs unless there is really different capability required.
- If the previous will take a long time to resolve, a DRS approach should be taken as there are immediate needs.  

- [ ] has_more
or
- [ ] page n of page_count

- [ ] page_count: number of pages (can we make this optional?)
what does its absence mean? If it means there is only one page, is it not easier to require page_count and set the value to 1 if that is the case.
- [ ] Navigation
next (URL to the next page
or to a specific page
previous (URL to the previous page) - seems less of a case for this

- [ ] items_per_page: number of bundle items per page

- [ ] Allow page size to be part of the request

bcli4d:
[...]

PR #367, opened 2021-09-14
DRS bundle contents pagination

jb-adams:
Issues: #366 , #325 

Built Docs: [link](https://ga4gh.github.io/data-repository-service-schemas/preview/feature/issue-366-bundle-contents-pagination/docs/)

This PR aims to specify pagination of the `contents` array for a single DRS Object. This is done by adding a `page` request query parameter, and a `Link` response header to inform the client of the traversal links to various pages.

Open Questions:
* A DRS Object has multiple array-type attributes. In addition to `contents` there is also `checksums`, `access_methods`, `aliases`. Do we foresee any of these attributes requiring pagination as well? If yes, we would likely need to rename the new parameter and header to avoid collision downstream (e.g. `Link` to `ContentsLink`)
* Does this introduction constitute a breaking change, since pagination is introduced on an existing endpoint that didn't have it previously? If so, we should think about how to get around this (such as another new query parameter that, when absent, causes the server to return the default, non-paginated response)

### DIFF
--- a/openapi/components/parameters/ContentsPage.yaml
+++ b/openapi/components/parameters/ContentsPage.yaml
@@ -1,0 +1,7 @@
+in: query
+name: page
+schema:
+  type: integer
+example: 5
+description: >-
+  For paginated traversal of the `contents` array of a DRS Object bundle. Indicates the requested "page" or subset of `contents` objects. 
--- a/openapi/components/responses/200OkDrsObject.yaml
+++ b/openapi/components/responses/200OkDrsObject.yaml
@@ -1,4 +1,54 @@
 description: The `DrsObject` was found successfully
+headers:
+  Link:
+    description: |
+      Pagination links to retrieve the `contents` array of a DRS Object bundle via paginated traversal.
+
+      #### Link header structure
+
+      The value of the `Link` header is a **comma-delimited** list of entries. Each entry in the list contains:
+      1. A URL that clients can request to obtain a new subset of `contents` for the same DRS Object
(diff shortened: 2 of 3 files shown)


ianfore:
See also https://github.com/ga4gh/data-repository-service-schemas/pull/377

jb-adams:
Following the Jan 10th Cloud WS call, I moved the paginated `contents` array function to 2 new endpoints:
- [GET /objects/{object_id}/contents](https://ga4gh.github.io/data-repository-service-schemas/preview/feature/issue-366-bundle-contents-pagination/docs/#operation/GetObjectContents)
- [POST /objects/{object_id}/contents](https://ga4gh.github.io/data-repository-service-schemas/preview/feature/issue-366-bundle-contents-pagination/docs/#operation/GetObjectContentsPost)

This was done to address the original awkwardness of having pagination for an attribute within the returned `DrsObject`. Now, the pagination affects the top-level array of the response. @patmagee @aniewielska 

We may need to back up the conversation a bit though, as @dglazer and @ianfore propose in #377  dropping bundles, and hence `contents` pagination, altogether. I'd prefer to arrive at an answer on whether bundles should be scrapped before doing more work in the branch.

@MichaelLukowski it would be really helpful to get your feedback on the U Chicago bundle use case to inform the conversation.

@briandoconnor

jb-adams:
Thanks for the feedback @MichaelLukowski ! Can you also relay your preference to drop bundles in thread #377 ? That's the PR that would remove bundles from the spec

Issue #378, opened 2022-01-14
Migrate CI to Github Actions


jb-adams:
As discussed at the Jan 10th call, we should migrate to Github Actions for several reasons, such as:
* establish common practices across Cloud API repos
* not having to use paid Travis CI plan

PR #379, opened 2022-01-18
migrate to GitHub actions

jb-adams:
Solution for #378 

Migrates CI processes (i.e. documentation build) to Github Actions from Travis CI.

Documentation: [link](https://jb-adams.github.io/data-repository-service-schemas/preview/support/issue-378-migrate-to-github-actions/docs/). There are no spec changes, this is only to show that the docs get built in the same way following the migration


aniewielska:
@jb-adams - a very nice CI. Where will I find the Dockerfile for `gh-openapi-docs`? Would it be possible to re-use the dockerized approach for `gh-openapi-docs` for other APIs?
Also I think you mentioned it, but would this CI actually work for both PRs from branch and PRs from fork? I assume, it will not for the latter due to [Github read-only permissions](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/) for `pull_request`, right?

Issue #353, opened 2021-04-28
DRS + Passports client role and capabilities


jb-adams:
A sub-problem of the [DRS + Passports Design Resolution](https://github.com/ga4gh/data-repository-service-schemas/issues/351) epic.

This sub-problem involves reaching consensus on what a DRS/Passport-aware client should be able to do to facilitate this flow. It also involves what information needs to be communicated to the client from the service.

Questions include:

- Where do we need to support handling passports that are too large for headers?
  - For some requests, an Authorization header may suffice, BUT we know some passport access tokens may be too large for headers
  - Should large passport tokens be passed in the request body and sent via POST request?
- How can we make the supported flows/capabilities self-describing in DRS?
  - If the DRS service supports OAuth2 bearer tokens, embedded passport access tokens, root passport access tokens, or some combination thereof, where is this communicated to client? Can/should this go in `service-info`?
- Can we support OAuth and passports?
  - Should we leverage an existing OAuth flow to do the token downscoping?
  - Can we self-describe so a client can discover which kind of flow a DRS server supports?

briandoconnor:
The Broad team reached out to me to ask about adding key/values to a DRS response to make it clear what auth mechanism to use for that resources.

mattions:
@briandoconnor this exactly the same problem I wanted to discuss in the Cloud Workstream meeting.

I think we need to add an authorization key to DRS, which points maybe to the iss, and provides type: ga4gh_passport maybe, so we can differentiate in case we have several DRS object, from the same server, with multiple authorization mechanism

PR #381, opened 2022-01-31
Add Authorizations information

dvoet:
This is an attempt to solve https://github.com/ga4gh/data-repository-service-schemas/issues/353.

The overall problem is how does a caller know how to authenticate for a given DRS URI and if passports are required, which to send?

This information cannot be added to the `server_info` endpoint because, depending on implementation, the answer may not be uniform across all objects serviced by a DRS provider.

`GetObject` and `GetAccessURL` may be protected differently (e.g. some implementations allow public access to `GetObject` but protect access to `GetAccessURL` while others protect both). This solution adds an `OPTIONS` api call to the `GetObject` url that returns information about how to authenticate for object resolution. The `AccessURL` was augmented to include the same information elements to indicate how to authenticate for `GetAccessURL`.


Avantol13:
> The overall problem is how does a caller know how to authenticate for a given DRS URI and if passports are required, which to send?

Is knowing which passport to send really feasible in a GA4GH world where passports can be repackaged (if the underlying information used to make an authorization decision is actually in the visa)? Sure there are cases where visas can't be repackaged (where knowing the Passport `iss` is enough), but there are also cases where perhaps multiple passport issuers could provide access to the data if they repackage the underlying visa that denotes access. 

I like the idea of a way for the client to know what types of auth a DRS server accepts (authHeader, Passport) but I'm not sure about implying that a specific passport `may` grant access based on `iss` is extensible (or even belongs in the DRS API). Nor am I sure that having this information per object is the best path given scalability, despite the added flexibility. 

My gut is telling me that there's some missing piece here, maybe a common policy language for authorization that the DRS server guarantees and is transparent to the client? But even that is too much bloat on DRS by itself in my opinion.

We have to consider the scalability impact of any solution that would require multiple calls from a client per individual DRS ID they want to access. The fact is that on average we're dealing with research workflows that access many many files over the course of a short amount of time. This bursty behavior would be 2x worse on DRS servers if the client had to make twice as many API calls.

I feel this is too much coupling of DRS with the concept of authorization when it really should be abstracted in the Clearinghouse functionality. DRS Servers by themselves shouldn't know if a given passport actually authorizes access to a given file. That's the job of the Clearinghouse. Perhaps we need to approach this from a different angle?

openapi/components/schemas/AuthorizationsProperties.yaml
```diff
@@ -0,0 +1,19 @@
...
+    type: string
+    enum:
+      - None
+      - BasicAuth
+      - BearerAuth
+      - PassportAuth
+  description: >-
+    An Optional list of support authorization types. More than one can be supported and tried in sequence.
+    Defaults to `None` if empty or missing.
+passport_issuers:
```
briandoconnor, line 13: Additionally, would it be useful/good to point to the passport broker URL?

jb-adams:
Could this addition potentially interfere with a [cross-origin-resource-sharing (CORS) preflight request](https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request)? The pre-flight makes use of an `OPTIONS` request at the same URL as a `GET` or `POST` to determine the server's CORS policy

openapi/components/schemas/AuthorizationsProperties.yaml
```diff
@@ -0,0 +1,19 @@
+authorizations:
+  type: array
```
briandoconnor, line 2: great to see this as an array since a given DRS object may have multiple auth approaches... e.g. Gen3 bearer token for dev access, RAS Passport for researcher access

briandoconnor:
Kurt can you help us get feedback on this?  @kwrodarmer

dvoet:
responses to @Avantol13 
> Is knowing which passport to send really feasible in a GA4GH world where passports can be repackaged (if the underlying information used to make an authorization decision is actually in the visa)? Sure there are cases where visas can't be repackaged (where knowing the Passport `iss` is enough), but there are also cases where perhaps multiple passport issuers could provide access to the data if they repackage the underlying visa that denotes access.

I thought we had given up on repackaging. If not, we could change this to visa issuer.

>I like the idea of a way for the client to know what types of auth a DRS server accepts (authHeader, Passport) but I'm not sure about implying that a specific passport may grant access based on iss is extensible (or even belongs in the DRS API). Nor am I sure that having this information per object is the best path given scalability, despite the added flexibility.

I think shortly we will live in a world with multiple passport providers granting access to different collections of data. Without understanding which passports to send to authorize a DRS request we are left with hardcoding based on some external knowledge (not possible in some cases) or just sending all the passports we have which is inappropriate. Not every DRS implementation deploys a new server per collection of data nor do I think we should require any notion of collection in the DRS spec. That leaves querying for authorization method per object as our only option.

>I feel this is too much coupling of DRS with the concept of authorization when it really should be abstracted in the Clearinghouse functionality. DRS Servers by themselves shouldn't know if a given passport actually authorizes access to a given file. That's the job of the Clearinghouse. Perhaps we need to approach this from a different angle?

Aren't DRS servers implementations of Clearinghouses? They are the things that [consume GA4GH claims to make an authorization decision](https://github.com/ga4gh/data-security/blob/master/AAI/AAIConnectProfile.md#term-claim-clearinghouse). I think it is desirable for the _GA4GH_ DRS spec to have strong ties to the _GA4GH_ passport spec.

On the scalability front, for Gen3 this can probably be a static response. In the more general case some providers could publish a caching strategy. We could also work some back pressure and retry strategies into the spec.

dvoet:
> Could this addition potentially interfere with a [cross-origin-resource-sharing (CORS) preflight request](https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request)? The pre-flight makes use of an `OPTIONS` request at the same URL as a `GET` or `POST` to determine the server's CORS policy

We can allow a 204 response to be treated the same as 405. This should allow this proposal to coexist with CORS (which uses headers instead of a response body). If required, the DRS OPTIONS api could return information in headers too.

openapi/components/schemas/AuthorizationsProperties.yaml
dvoet: I think this is hard. Each consuming system of DRS likely has its own Broker (e.g. Terra has its own, SBG probably does too). How would the DRS server know the list? I think where the caller gets their passport should stay out of this spec. Said another way, I think this spec should be able to say what passports it needs but not where or how to get them.

kwrodarmer:
@dvoet  - the DRS server provider does have it within their facile control to provide different servers based not upon data set but upon governing authority behind the data set. In this way, a fully-specified DRS id resolved down to the host does give some indication of what token is needed for authorization to access hosted content.

kwrodarmer:
@dvoet - our DRS server makes use of separate services for its clearinghouse. From the caller's perspective it does perform the clearinghouse functionality. From the implementation perspective, the core clearinghouse functionality is implemented centrally (the hub) within the organization and the query evaluation is performed in spokes in our case.

mattions:
This PR goes into the right direction.

In general this will work similar to:

```
# call is OPTIONS because the first GET _can_ be controlled. 
# 
OPTIONS /ga4gh/drs/v1/drs.example/a32131sadsaeq32 

# The response is something on the line
{
"authorizations" : ["PassportAuth"]
"visa_issuers: ["https://stsstg.nih.gov/"]
}
```
# where the iss is exactly the url in the VISA, for example that  https://auth.nih.gov/docs/RAS/serviceofferings.html#AppendixA

One question I have is, would it make sense to have the `visa_issuer` renamed to `authorization_issuer` so we could expand that also for the Authentication Bearer case?

It would look like this for the Passport case:
```
{
"authorizations" : ["PassportAuth"]
"authorizaion_issuers: ["https://stsstg.nih.gov/"]
}
```
and this for the Bearer Token.

```
{
   "authorizations" : ["BearerAuth"]
   "authorization_issuers: ["https://example.com/"]
}
```

@dvoet what do you think?

dvoet:
@mattions I am a little concerned about broadening the `visa_issuers` field to cover `BearerAuth` because it opens an attack vector where a service could request bearer tokens for other services. I agree that a similar attack is possible for passports but that is mitigated by mTLS and passports only grant access to data and not other functions. I also feel like it is an already solved problem because if a service can generate access tokens to access some other service, it already should know what resource servers accept those access tokens (in other words, the url of the resource server should be enough to figure out which bearer token to use and we don't need another field to tell us that).

mattions:
@dvoet I do not think we are talking about the same problem.

We have use-case where a DRS Server host data with different type of authorization. Having the ability to identify which is the authorization issuer, provides the ability to the DRS Client connecting to that Server the ability to know which authorization they have to send.

dvoet:
@mattions if you have a use case where the same resource server will accept more than one bearer token issuer then I am fine with the change. I will also add language that it is recommended to independently verify that it is appropriate to send the requested credentials.

mattions:
@dvoet I think that would be a nice addition to the specifications, which will provide us to specify which authorization schema a DRS client should follow.

@briandoconnor @jb-adams I would like to understand what needs to be done to have this PR move forward.

jb-adams:
> @dvoet I think that would be a nice addition to the specifications, which will provide us to specify which authorization schema a DRS client should follow.
> 
> @briandoconnor @jb-adams I would like to understand what needs to be done to have this PR move forward.

@mattions I think the main thing here is that this proposal aligns with the in-development Passports v1.2 spec (downscoping, work order tokens, etc). @kwrodarmer what are your thoughts on this?

If there's alignment between this and Passports then @briandoconnor and I can bring this up at a future Cloud WS call, which will speed up the review/merging process. We should also aim to have this implemented somewhere before we merge as a demonstration that it is feasible (I could work on this on my reference implementation).

ianfore:
Could anyone help reconcile this with the implementation in [GA4GH Starter Kit DRS](https://github.com/ga4gh/ga4gh-starter-kit-drs)?

An OPTIONS call to 
http://localhost:5000/ga4gh/drs/v1/objects/28ffe6386b78fffb7cd0a7b2d0a79846
gives
`{
    "supported_types": [
        "PassportAuth"
    ],
    "passport_auth_issuers": [
        {
            "broker_url": "http://localhost:4455/",
            "visa_name": "1000GenomesIndividualsWithAmericanAncestry",
            "visa_issuer": "https://federatedgenomics.org/"
        }
    ]
}`

Should we clarify with StarterKit team whether in the course of implementing they identified additional needs that resulted in this form of response?

Issue #337, opened 2020-11-10
Bundling and other approaches to mapping accessions/logical entities to DRS/ physical objects


ianfore:
There is a need to explore the use of bundling as an approach to dealing with complex objects in DRS.

Cross referencing to to https://github.com/ga4gh/cloud-interop-testing/issues/104 which has the details of a worked example using SRA.

ianfore:
The difference in approaches has been illusrated in this Jupyter notebook using an SRA example.

An imaging example has been used to illustrate maintaining the logical structure of collections of bundles in an external model. This allows DRS to remain a simple foundational protocol for access to physical objects in the cloud.

PR #389, opened 2022-08-12
deprecating bundles (documentation changes only)

dglazer:
Per several discussions at our weekly WG meetings, and discussions with driver projects, this PR updates the documentation to say that bundles are deprecated, and to describe best practices for working with simple and compound datatypes in a post-bundle world. (Once merged we'll be able to close out a few open issues around improving bundle support.)

Preview the built docs at https://ga4gh.github.io/data-repository-service-schemas/preview/feature/issue-334-bulk-requests-nobundle-v2/docs/#section/DRS-Datatypes

### DIFF
--- a/openapi/tags/DrsApiPrinciples.md
+++ b/openapi/tags/DrsApiPrinciples.md
@@ -110,10 +110,10 @@ DRS's job is data access, period. Therefore, the DRS API supports a simple flat 
 DRS can be used to access individual objects of all kinds, simple or complex, large or small, stored in type-specific formats (e.g. BAM files, VCF files, CSV files). At the API level these are all the same; at the application level, DRS clients and servers are expected to agree on object semantics using non-DRS mechanisms, including but not limited to the GA4GH Data Connect API.
 
 ### Compound Objects
-DRS can also be used to access compound objects, consisting of two or more atomic objects related to each other in a well-specified way. See the [Appendix: Compound Objects](#tag/Compound-Objects) for suggested best practices for working with Compound Objects. 
+DRS can also be used to access compound objects, consisting of two or more atomic objects related to each other in a well-specified way. See the [Appendix: Compound Objects](#tag/Working-With-Compound-Objects) for suggested best practices for working with Compound Objects. 
 
 ### [DEPRECATED] Bundles
-Previous versions of the DRS API spec included support for a *bundle* content type, which was a folder-like collection of other DRS objects (either blobs or bundles), represented by a `DrsObject` with a `contents` array. As of v1.3, bundles have been deprecated in favor of the best practices documented in the [Appendix: Compound Objects](#tag/Compound-Objects). A future version of the API spec may remove bundle support entirely.
+Previous versions of the DRS API spec included support for a *bundle* content type, which was a folder-like collection of other DRS objects (either blobs or bundles), represented by a `DrsObject` with a `contents` array. As of v1.3, bundles have been deprecated in favor of the best practices documented in the [Appendix: Compound Objects](#tag/Working-With-Compound-Objects). A future version of the API spec may remove bundle support entirely.
 
 ## Read-only
 
--- a/openapi/tags/DrsApiPrinciples.md
+++ b/openapi/tags/DrsApiPrinciples.md
@@ -110,10 +110,10 @@ DRS's job is data access, period. Therefore, the DRS API supports a simple flat 
 DRS can be used to access individual objects of all kinds, simple or complex, large or small, stored in type-specific formats (e.g. BAM files, VCF files, CSV files). At the API level these are all the same; at the application level, DRS clients and servers are expected to agree on object semantics using non-DRS mechanisms, including but not limited to the GA4GH Data Connect API.
 
 ### Compound Objects
-DRS can also be used to access compound objects, consisting of two or more atomic objects related to each other in a well-specified way. See the [Appendix: Compound Objects](#tag/Working-With-Compound-Objects) for suggested best practices for working with compound objects. 
+DRS can also be used to access compound objects, consisting of two or more atomic objects related to each other in a well-specified way. See the [Appendix: Compound Objects](#tag/Working-With-Compound-Objects) for suggested best practices for working with compound objects.
 
 ### [DEPRECATED] Bundles
-Previous versions of the DRS API spec included support for a *bundle* content type, which was a folder-like collection of other DRS objects (either blobs or bundles), represented by a `DrsObject` with a `contents` array. As of v1.3, bundles have been deprecated in favor of the best practices documented in the [Appendix: Compound Objects](#tag/Working-With-Compound-Objects). A future version of the API spec may remove bundle support entirely.
+Previous versions of the DRS API spec included support for a *bundle* content type, which was a folder-like collection of other DRS objects (either blobs or bundles), represented by a `DrsObject` with a `contents` array. As of v1.3, bundles have been deprecated in favor of the best practices documented in the [Appendix: Compound Objects](#tag/Working-With-Compound-Objects). A future version of the API spec may remove bundle support entirely and/or replace bundles with a scalable approach based on the needs of our driver projects.
 
 ## Read-only
 
--- a/openapi/tags/DrsApiPrinciples.md
+++ b/openapi/tags/DrsApiPrinciples.md
@@ -99,7 +99,7 @@ DRS servers can choose to issue either hostname-based or compact identifier-base
 
 |                   | Hostname-based | Compact Identifier-based |
 |-------------------|----------------|--------------------------|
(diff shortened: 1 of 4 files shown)


kwrodarmer:
The deprecation of bundles affects dbGaP, and is something I'm not entirely comfortable with.

What I have said before is that I support moving from bundle as a type of substitute for what could otherwise be described as a set, but there is still utility in being able to designate structured objects such as SRA runs. I will talk more with @ianfore about this, but currently dbGaP and the SRA depend upon the ability to translate an SRA accession into a DRS bundle.

mattions:
Shall we try to have a discussion about this during the plenary, so we can better figure out what to do about it?

ianfore:
A discussion would be good.
Suggested reading and understanding the SRA example here would be excellent preparation for such a discussion.
https://github.com/ga4gh/fasp-scripts/blob/master/notebooks/FASPNotebook14-SRAExample.ipynb

Revisiting how we got to this PR, see issue https://github.com/ga4gh/data-repository-service-schemas/issues/337
In a sense nothing new beyond the example above. Only referring to it to show the SRA use case was part of the consideration.

bcli4d:
At which plenary session will this be discussed?

ianfore:
@bcli4d - the cloud session on 21-Sep I would think. [Agenda here](https://docs.google.com/document/d/16zw509A99xiRSBIbb7oGxFGxiNGgBc0eqY060pSwzAc/edit#heading=h.ihrn1svd5d71). Probably within the DRS item. 

May also need more detailed work as part of the 3 day Connect meeting (Oct 19-21).

fedorov:
NCI Imaging Data Commons (IDC) (https://imaging.datacommons.cancer.gov) has an urgent need to support access to its data via some sort of bundling mechanism. We use DICOM for harmonizing our data, and in DICOM, in most cases, each individual slice is represented as a file on disk. Since access to the data is most often done at the granularity of DICOM series, which often contain hundreds of slices and in turn files, analysis cohorts defined as lists of files become unnecessarily large.

We would use the DRS bundles object for this purpose as it is a reasonable fit to our use case and it is used, at least to some degree, by Seven Bridges and other entities of NCI Cancer Research Data Commons. However, the DRS bundle design apparently presents serious implementation issues.

Therefore, if Cloud WS can "fix" the bundle definition and if DCF can implement such a fix, all in a timely manner, we will use bundles. Otherwise, we will work within CRDC on a CRDC-defined standard complex object that can provide functionality similar to bundles.

We, the undersigned, on behalf of the IDC team: @fedorov @bcli4d @wlongabaugh

ianfore:
Thanks Andrei. Yes a long running story on this. Was going to revisit some of the details but a different possibility just occurred to me. What if the DRS Object was a folder. See #391 for discussion as it would distract from this thread. 

And even, typing the DRS objects is desirable. application/dicom is a registered mime type with IANA.

dglazer:
@fedorov -- thank you for the thoughtful response around how to use DRS for your DICOM needs. DICOM was one of the original motivating use cases for this compound object proposal, so I'm glad you think it will work for you. I continue to believe that will be a better solution than stretching bundles, because a) I think it provides cleaner ways to handle scalability, and b) because it lets you define the semantics of the compound object to meet your needs, which are different than those of other compound object types. We can discuss more at or after the upcoming Cloud session.

dglazer:
@kwrodarmer -- I agree that DRS needs to support structured objects like SRA runs and accessions. I believe defining a compound object will be a better way to do that than trying to fit everything into a generic bundle. Curious to hear how your discussion with @ianfore went, and/or to continue it at the Cloud WS session and beyond.

ianfore:
@fedorov -- I agree with @dglazer that defining the compound object manifest fits with the proposed change. It's not so much an 'otherwise' as exactly what the David intended in the PR. 

I believe you have already this in place. It seems the semantics/model likely to be needed are provided by the IDC BigQuery tables. I worked up an example that gets the DRS ids for a series from BigQuery and then accesses the images via DRS as needed in a viewer. Happy to demonstrate as soon as we can set it up.
 
So far I have used dicom_all, but it could be one of the other tables would be more efficient. Perhaps you could help with this?

The bulk request option in DRS 1.3 is certainly needed to support this - but that doesn't seem to be at issue.

On one note:
> Since access to the data is most often done at the granularity of DICOM series, which often contain hundreds of slices and in turn files, analysis cohorts defined as lists of files become unnecessarily large.

At the series level the bundle is a list of files. I'm not sure the original bundle helps with the issue you raised. Could someone illustrate? Again bulk request is the useful addition.

Also. We've seen in other cases that asynchronous download of files, followed by processing, isn't the most efficient. For a viewer use case, series of any significant size are not be viewable at all once. This gives the option for immediate access to what will be seen, while the remainder of the series is downloaded synchronously.

bcli4d:
@dglazer, in reviewing the proposed description of compound objects, etc., I'm concerned that someone might conclude that a compound object cannot reference other compound objects. The text kind of equates 'atomic object' with something like a file that does not reference other objects. E.G. "Much content (e.g. VCF files) is well represented as a single atomic `DrsObject`." 
 
It then seems to describe a compound object as only referencing atomic objects. E.G. "Some content, however (e.g. DICOM images) is best represented as a compound object consisting of a structured collection of atomic `DrsObject`s." and "DRS can also be used to access compound objects, consisting of two or more atomic objects related to each other in a well-specified way."

Also, there is this:
"### Atomic Objects
DRS can be used to access individual objects of all kinds, simple or complex, large or small, stored in type-specific formats (e.g. BAM files, VCF files, CSV files). At the API level these are all the same; at the application level, DRS clients and servers are expected to agree on object semantics using non-DRS mechanisms, including but not limited to the GA4GH Data Connect API."
Is a "complex" object the same as a compound object. I'm guessing you do not mean that, but "complex object" and "compound object" have been used interchangeably in some of these discussions. 

I think we should state  that an atomic object is one which does not contain DRS URIs (if that is how we are to understand it :-) and state explicitly that a compound object can reference both atomic objects and other compound objects.

Bill

dglazer:
Good catch Bill -- I'll let you know when I've clarified the docs.

briandoconnor:
@dglazer were you able to clarify the docs based on Bill's comments?

To keep this on track for Connect in April 2023 can you confirm the following:

Champion → David G?
Deadline for revisions: ?
Deadline for comments: ? 
Merge date: ? (3/13 at latest)

openapi/tags/DrsApiPrinciples.md
```diff
@@ -104,11 +104,16 @@ DRS servers can choose to issue either hostname-based or compact identifier-base
...
 
-* a *blob* is like a file — it’s a single blob of bytes, represented by a `DrsObject` without a `contents` array
-* a *bundle* is like a folder — it’s a collection of other DRS content (either blobs or bundles), represented by a `DrsObject` with a `contents` array
+### Compound Objects
+DRS can also be used to access compound objects, consisting of two or more atomic objects related to each other in a well-specified way. See the [Appendix: Compound Objects](#tag/Working-With-Compound-Objects) for suggested best practices for working with compound objects. 
+
+### [DEPRECATED] Bundles
+Previous versions of the DRS API spec included support for a *bundle* content type, which was a folder-like collection of other DRS objects (either blobs or bundles), represented by a `DrsObject` with a `contents` array. As of v1.3, bundles have been deprecated in favor of the best practices documented in the [Appendix: Compound Objects](#tag/Working-With-Compound-Objects). A future version of the API spec may remove bundle support entirely.
```
briandoconnor, line 116: I would like to change "A future version of the API spec may remove bundle support entirely." to "A future version of the API spec may remove bundle support entirely and/or replace bundles with a scalable approach based on the needs of our driver projects."  The reason I want to add this is I don't think we can just say "use compound objects, the semantics of which are beyond our definition of this spec" since it just means everyone does something different.  I'm hearing from drivers that the concept of bundles is useful but they don't scale (e.g. dbGaP, gen3, etc).  And until we have something better that scales, the compound object approach with manifests gives folks a scalable way to work today with the current spec (great!).  This PR gives us a best practice when scale matters.  But I don't want to cut us off from a solution that involves introduction of a scalable set endpoint for example.

briandoconnor (changes requested):
See my feedback for a small change to leave things open for a better bundle replacement in the future if our drivers need this.

openapi/tags/DrsApiPrinciples.md
briandoconnor: We're slated to explore alternatives during the April 2023 Connect meeting.  See the [DRS and Data Connect](https://docs.google.com/document/d/1s5aPcNQmwmin1c5F1la4O5LJDDz9VB0p/edit) session.  

Since there has been no additional comments on this since our last DRS-focused cloud meeting on 1/23/23 I'm assuming everyone is fine with moving this forward (including the small tweak above).  I'll merge into development.

If you are a stakeholder in this topic, please bring your ideas/opinions to the above GA4GH Connect session, we will target having a viable bundles replacement during that meeting.  And will work in future versions of this call to update DRS 1.4 accordingly.

briandoconnor (approved):
I made a small change, see the comment thread.  Otherwise LGTM.  It's been 2 months since the last conversation on this so I'm going to move the review forward and merge with develop.  If there's any future feedback, we can always create a new issue/PR to address before the 1.4 release.

Issue #343, opened 2021-01-26
DRS and Requester Pays


briandoconnor:
I'm making this ticket to trigger conversation around what changes (if any) are needed with regards to supporting signed URLs on requester pays objects in cloud storage.

ianfore:
In today's FASP hackathon topic of enabling "get a url" to support use of "requester pays" buckets 

Some of the requirements for this were explored in accessing Anvil data through DRS. A workaround was implemented in  [this anvilDRSClient](https://github.com/ga4gh/fasp-scripts/blob/ec4b9caf835a355a72675b57e3ab174404baddee/fasp/loc/gen3drsclient.py#L146) class.

It was a relatively easy fix in that the URL provided by the Anvil DRS simply needed to have &userProject=_projectid_ appended to it.

For convenience in the class above the user project was provided at class instantiation, but it might equally be applied at the method level i.e.
def getAccessURL(self, object_id, access_id=None, user_project=None)

If implemented directly in the DRS API the additional parameter of user_project would likely be sufficient.

**Some caveats**
- This still requires a change to the privileges on userProject to allow the Google account used by the DRS service to bill to userProject. See comment in the class for the specific IAM privilege required. Making such a configuration change to their project would not be simple for many users.
- The above is specific to Google Cloud Storage. Other platforms may have other requirements.
- The hope is that in general users would not be downloading and therefore not incurring the egress charges. However, Google still requires a userProject to be provided in the URL even if no billing will occur. TBH a change to this by Google would be preferable. This has been raised at STRIDES Google Office Hours and could be revisited.

patmagee:
@ianfore the other option would be going in the opposite direction and define a DRS access method that could accept any additional credential and parameter needed to generate a signed url that bills the appropriate entity. This would necessitate that we allow a `POST` access-method, not just a `GET` 

The problem with requester pays in general, is that any way we cut it, we need to either trust the drs server with a credential (Service account) or provide them with billing usage of our own personal project. Both approaches have their pro's and con's, but I personally shy away from explicitly granting any service permissions directly within my project. The other approach would be for the user to provide the signing credentials. Here there is implicit trust that the DRS server never re-uses the signing credentials that are passed in the request, which I find a more palpable trust model.

So what does this look like practically?
1. There is a `POST access-method` for a given drs object
2. This access method accepts the `user_project` param and a `service_account` json object
[...]

Issue #336, opened 2020-11-03
adding/standardizing field providing reference to external metadata


bheavner:
The BioData Catalyst Data Harmonization Working Group would like to use the key:value functionality of DRS to provide a link to external metadata services. For example, we might want to include a URI or list of URIs for external metadata service(s) (e.g. API endpoint) where one could obtain additional metadata for an object by querying with a drs guid.

We recognize that we can implement this solution in a platform-specific way, but perhaps if we were to standardize on the field name and allowable/required data types and values, it would facilitate future interoperability of tools that use such a "metadata" field.

mattions:
Thanks @bheavner . Just to get started the conversation, I think a possible addition with a "dictionary format" could be a starting point.

So to the standard DRS reply 
```
{
  "id": "string",
  "name": "string",
  "self_uri": "drs://drs.example.org/314159",
  "size": 0,
  "created_time": "2020-11-05T14:52:31.054Z",
  "updated_time": "2020-11-05T14:52:31.054Z",
  "version": "string",
  "mime_type": "application/json",
  "checksums": [
    {
      "checksum": "string",
      "type": "sha-256"
    }
  ],
  "access_methods": [
    {
      "type": "s3",
      "access_url": {
        "url": "string",
        "headers": {
          "Authorization": "Basic Z2E0Z2g6ZHJz"
        }
      },
      "access_id": "string",
      "region": "us-east-1"
    }
  ],
  "contents": [
    {
      "name": "string",
      "id": "string",
      "drs_uri": "drs://drs.example.org/314159",
      "contents": [
        {}
      ]
    }
  ],
  "description": "string",
  "aliases": [
    "string"
  ]
}
```
we could add a field that could have 
```metadata : {"type": "<API_type>", "uri" : "<url_to_the_server>"}```

this could also provide a way to have more than one pointer (if needed) and different ways to point to the metadata.

bheavner:
That seems like a very useful way to go, thanks for the suggestion!

Would <API_type> be enumerated, or free text? I'm not sure what types of API may exist or be desirable, or what values would be used for that field. For example, I can imagine someone using something like swagger to roll their own, or there may be some domain-specific metadata endpoints that I'm not aware of. I can imagine someone having metadata available as a html page or free text, which would likely be a dead-end for standard downstream use, but would still be good to be able to point to.

mattions:
[...]

PR #390, opened 2022-08-19
Feature/issue 343 metadata endpoint

MichaelLukowski:
This is a PR for introducing a new metadata endpoint that supports an arbitrary JSON response at a new `/metadata/{object_id}` endpoint.

There is a need for requestor pays and egress for additional metadata of DRS objects that shouldn't be included in the `/object` endpoint.

This will also be useful for a best practice of "complex objects" where one DRS id can point to a json file that describes many DRS objects.


dglazer:
Sorry if I missed earlier discussion of this PR, but I'm not following the intended use of this new method, or how it would address the stated goals fetching requestor-pays content or supporting complex objects. Is there more background somewhere? (I did look at #343 and didn't see answers there.) I'll reserve most comments until I hear back, but in general I think providing a method that returns arbitrary unspecified metadata creates a big risk of breaking interoperability.

ianfore:
Referring back to this [hackathon exploration](https://github.com/ga4gh/fasp-scripts/blob/master/notebooks/drs/altMetadata.ipynb) and [the DRS issue](https://github.com/ga4gh/data-repository-service-schemas/issues/336) that generated it.

It may be a separate use case - all depends on exactly what metadata are intended in each situation. The hackathon was following an approach that aligns with "bundling alternatives" described in PR 389. Both 336 and 389 rely on higher level schemas that provide the metadata.

Despite the ambiguity, the motivation of this issue and 389 seems to be the same. So am also waiting on what @dglazer asked.

Commenting on 
`arbitrary unspecified metadata creates a big risk of breaking interoperability.`
Yes, a shared concern. A key characteristic of the protocols (e.g. Data Connect)  that provide the higher level schemas is that allow for, and encourage, for those metadata to be well specified.

briandoconnor:
@MichaelLukowski just wanted to check in with you on this PR

To keep this on track for Connect in April 2023 can you confirm the following:

Champion → @MichaelLukowski 
Deadline for revisions: ?
Deadline for comments: ?
Merge date: ? (3/13 at latest)

briandoconnor:
@MichaelLukowski suggested this might be closed and replaced with GA4GH Connect... since it already provides the ability to specify a metadata definition.

briandoconnor:
https://github.com/ga4gh-discovery/data-connect

briandoconnor:
@ianfore has an example that we can point to... from FASP work.  Ian can you drop links here?

briandoconnor (changes requested):
Based on Cloud call on 1/23/23 it sounds like Michale wants to withdraw this PR and recommend using Data Connect.

@MichaelLukowski can you document this decision here and point to data connect, maybe an example of how to do a call to data connect to get back DRS URIs + metadata

briandoconnor:
Examples from @ianfore 
Examples:
DRS and metadata exploration
https://github.com/ga4gh/fasp-scripts/blob/master/notebooks/drs/altMetadata.ipynb
https://github.com/ga4gh/fasp-scripts/tree/master/notebooks/drs

Issue #395, opened 2023-07-13
Support for data is in cold storage


ianfore:
It has been suggested we work on support for cold storage in the DRS specification. Ahead of submitting a pull request for this. It seems worth laying out some of the considerations for support of cold storage. Not all considerations need be accommodated, but it helps to have an idea of the overall landscape.

'Cold' storage is a shorthand for the situation where an object is not immediately available to a 'get a URL' DRs request ( a request of the form /objects/<drs_id>/access/<access_id>). Hot/Cold is binary. The different storage tiers offered by many providers have more subtle gradations of availability. However for this discussion we will assume that hot/cold is sufficient, unless someone suggests it needs to be more complex. 

- Indicate that a given **access method** is to cold/offline/less available storage
A given DRS Object may have access methods for both hot and cold storage i.e. the data model is that hot/cold is an attribute of an Access Method not of a DRS Object.

e.g. something like the following pseudo-specification is likely necessary
```
{
   "access_methods": [
      {
         "access_id": "e93724",
         "region": "ncbi",
         "type": "https"
      },
      {
         "access_id": "fbd466",
         "region": "gs.US",
         "type": "https",
         "storage": "cold"
      },
      {
         "access_id": "0da151",
         "region": "s3.us-east-1",
         "type": "https"
      }
   ],
   "checksums": [
      {
         "checksum": "044e759c2e430c3db049392b181f6f5a",
         "type": "md5"
      }
   ],
   "created_time": "2022-06-03T14:07:27Z",
   "id": "044e759c2e430c3db049392b181f6f5a",
   "name": "SRR000066.lite",
   "self_url": "drs://locate.ncbi.nlm.nih.gov/044e759c2e430c3db049392b181f6f5a",
   "size": 118588310
}
```

- DRS should provide the mechanism for a client to request that the copy of the object at a specified access method be 'thawed'.

- Having asked for an object to be thawed, the client should be notified, or be able to check, whether the object is now available in immediately available (hot) storage.
 
- At the point when the object is available hot, the client would obtain the URL using the existing DRS method to get the URL.
 
- All the above should be possible in 'bulk mode' i.e. a client would want to make a single request for some large number of objects to be thawed. One likely use case is that a whole study or collection is in cold storage, and one would want to request all objects.
As with other bulk approaches DRS would concern itself only with lists of objects to be thawed - not what the external entity, like study, that those lists represent.

[...]

PR #401, opened 2023-12-18
Feature/issue 395 Indicate object in cold storage

suryasaha:
Indicate that a given access method is to cold/offline/less available storage

A DRS object is said to be in 'Cold' storage when it is not immediately available via a DRS request. Hot/Cold is binary. Goal is to indicate that a given access method is to a cold/offline/less available storage. Please see https://github.com/ga4gh/data-repository-service-schemas/issues/395 for community discussion.

We are taking a phased approach. For Phase 1, we just want to answer this question:
Can the user can retrieve the file at the moment of the request?
 - If the user can do that, than the file is in "hot" storage, and can be used right away.
 - If the user cannot retrieve the file, then the file is simply in not available. It is in some form of cold/offline/less available storage.

In phase 2, we can think about having a better way to provide details on the storage and the thawing process.

### DIFF
--- a/openapi/components/schemas/AccessMethod.yaml
+++ b/openapi/components/schemas/AccessMethod.yaml
@@ -29,12 +29,12 @@ properties:
     description: >-
       Name of the region in the cloud service provider that the object belongs to.
     example: us-east-1
-  storage:
-    type: string
+  availability:
+    type: boolean
     description: >-
-      Type of storage of file in the cloud.
-      Cold defines that this file is not accessible as its in offline/archival storage.
-    example: cold
+      Availablity of file in the cloud.
+      This label defines if this file is immediately accessible via DRS. Any delay or requirement of thawing mechanism if the file is in offline/archival storage is classified as unavailable.
+    example: true
   authorizations:
     allOf:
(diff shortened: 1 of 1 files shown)


briandoconnor:
See cloud work stream discussion notes from last DRS discussion from 2/12/24: https://docs.google.com/document/d/1hayvWLIoymomPI9oXcaTZirn5YxFv1cYAs70zyvlvnA/edit#bookmark=id.stgpsb1mhzwn

mattions:
quick note form Connect Switzerland, recording the decision:

Let's make it an enumeration: --> immediate, delayed as the two terms.
Lots of thumbs up in the room.

suryasaha:
Interest from driver projects at Connect: CRDC, BDC, SRA, European 1M genomes project, AnVIL, Australian Genomics

suryasaha:
@ianfore I am using _availability_ as the label now with boolean value. Do you think this looks better?

briandoconnor:
as of 20240812 we decided the boolean is the right way to go.  We can discuss delay time reporting back during the thaw mechanism

Issue #400, opened 2023-12-05
locations constraints on DRS Pointer


mattions:
In CRDC driver Project and also in BioDataCatalyst we have a situation where the host of the data would like to provide a guidance on how to use the data, and there to use it.

In other words, they would like that any platform downstream of the DRS Server would compute on the data in certain cloud locations, which usually are the same where the data are from. The reasons for this request are different, going from keeping the egress cost down, to not having the data leaving the security level.

Given that at the end we have download url in DRS, and it would be pretty difficult to enforce the situation, therefore I suggest we go more towards an idea where the host "suggest" what is the preferred way to access the data, and the DRS client accessing these data honor the request to the best of their ability.

# Proposal
The proposal aims to enhance the GA4GH DRS (Data Repository Service) specification by introducing a new field that provides metadata regarding the intended usage and location constraints for data objects. This additional field will allow data providers to specify their preferences and requirements for how the data should be accessed and utilized. The proposed field will offer the following options: 

1. **Cloud Exclusive** (_cloud_exclusive_): the data object is intended for use **exclusively within a cloud environment**. Users are expected to access and process the data only within a cloud computing infrastructure and not outside of it; cannot download the data on somebody's laptop 

2. **Cloud Provider-Limited** (_cloud_provider_limited_): the data object **should not leave the cloud provider's ecosystem**. Users are restricted from moving the data to external locations or platforms. It must remain within the boundaries of the specified cloud provider. 

3. **Cloud Region-Limited** (_cloud_region_limited_): the data object **is restricted to a specific cloud region**. Users are required to access and process the data within the designated region and are prohibited from transferring it to other geographic locations within the cloud provider's infrastructure.  

By introducing this new field, data providers and administrators can communicate their data access and usage policies more effectively, ensuring that data is handled in accordance with their specific requirements. This addition not only enhances the flexibility of the DRS specification but also strengthens data governance and compliance for genomic and health-related data in cloud-based environments. 

 It could look like this:
```
{ 
  "id": "string", 
  "name": "string", 
  "self_uri": "drs://drs.example.org/314159", 
  "size": 1024, 
  "created_time": "2019-08-24T14:15:22Z", 
  "updated_time": "2019-08-24T14:15:22Z", 
  "version": "string", 
  "mime_type": "application/json", 
  "checksums": [ 
    { 
      "checksum": "string", 
      "type": "sha-256" 
    } 
  ], 
  "usage_constraints": { 
[...]

PR #405, opened 2024-04-22
Feature/issue 400 location constraints

jared-rozowsky:
usage_constraints is an optional field that specifies how the data owner intends for the data to be used on the cloud.

This spec aims to enhance the GA4GH DRS (Data Repository Service) specification by introducing a new field that provides metadata regarding the intended usage and location constraints for data objects. This additional field will allow data providers to specify their preferences and requirements for how the data should be accessed and utilized. The proposed field will offer the following options:

- Cloud Exclusive (cloud_exclusive): the data object is intended for use exclusively within a cloud environment. Users are expected to access and process the data only within a cloud computing infrastructure and not outside of it; cannot download the data on somebody's laptop

- Cloud Provider-Limited (cloud_provider_limited): the data object should not leave the cloud provider's ecosystem. Users are restricted from moving the data to external locations or platforms. It must remain within the boundaries of the specified cloud provider.

- Cloud Region-Limited (cloud_region_limited): the data object is restricted to a specific cloud region. Users are required to access and process the data within the designated region and are prohibited from transferring it to other geographic locations within the cloud provider's infrastructure.

By introducing this new field, data providers and administrators can communicate their data access and usage policies more effectively, ensuring that data is handled in accordance with their specific requirements. This addition not only enhances the flexibility of the DRS specification but also strengthens data governance and compliance for genomic and health-related data in cloud-based environments.

### DIFF
--- a/openapi/components/schemas/AccessMethod.yaml
+++ b/openapi/components/schemas/AccessMethod.yaml
@@ -34,7 +34,7 @@ properties:
       - $ref: './Authorizations.yaml'
       - description: When `access_id` is provided, `authorizations` provides information about how to authorize the `/access` method.
   usage_constraints:
-    type: array
+    type: object
     items:
       - $ref: './UsageConstraints.yaml'
       - description: >-
--- a/openapi/components/schemas/AccessMethod.yaml
+++ b/openapi/components/schemas/AccessMethod.yaml
@@ -33,3 +33,10 @@ properties:
     allOf:
       - $ref: './Authorizations.yaml'
       - description: When `access_id` is provided, `authorizations` provides information about how to authorize the `/access` method.
+  usage_constraints:
+	  type: array
+	  items:
+	  	$ref: ‘./UsageConstraints.yaml’
+	  description: >-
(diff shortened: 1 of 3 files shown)


briandoconnor:
In the Cloud WS meeting on [Aug 12th, 2024](https://docs.google.com/document/d/1hayvWLIoymomPI9oXcaTZirn5YxFv1cYAs70zyvlvnA/edit#bookmark=id.jg4lev1aum29) we decided to simplify the feature described in [Issue #400](https://github.com/ga4gh/data-repository-service-schemas/issues/400) for DRS release 1.5.

PR #407, intended for DRS release 1.5, simply adds a string "cloud" to the access response.  We now include cloud, region, and type information only… no cloud or geo location constraint support for example.

The fields we will include are:
- Cloud: e.g. gcp or aws or azure (this is the new field)
- Type: s3 or https or etc
- Region: e.g. us-east-1

After DRS 1.5 we can revisit how we express region, cloud, geo location, etc constraints in DRS which is a much bigger issue.

Issue #394, opened 2023-05-22
Best practice for using DRS and Data Connect together


briandoconnor:
### Background

I'm opening this issue based on followup to the April 20th, 2023 GA4GH Connect meeting "[DRS and Data Connect](https://docs.google.com/document/d/1oLvgyvxAk_q3arRk0JaHbWAhiK54bkvR/edit#)" session.  This session looked at exploring how standards from the Cloud and Discovery work streams can be used together to identify the two needs identified in the aims listed below:
* Address the need to obtain additional data about a DRS object
* Revisit how Data Connect handles the need for bundles

Some resources of interest:
* Original issue for extra metadata in DRS: https://github.com/ga4gh/data-repository-service-schemas/issues/336
* Pull request - intent is to close, still has useful for review of requirements: https://github.com/ga4gh/data-repository-service-schemas/pull/390 
* DRS and metadata exploration to revisit
  * https://github.com/ga4gh/fasp-scripts/blob/master/notebooks/drs/altMetadata.ipynb 
  * https://github.com/ga4gh/fasp-scripts/tree/master/notebooks/drs

#### Metadata + DRS
We agreed that best practices for working with metadata were important, and largely agreed on two guiding principles:
* 1) DRS doesn’t know about metadata, and shouldn’t. Instead, we should lean into the fact that systems that use DRS typically have some database-like component that does know about object metadata. 
* 2) No new APIs (or API changes for DRS) are needed. Instead, we should add an appendix to the DRS spec documenting best practices for building systems that use DRS and care about metadata.

#### Compound Objects
We agreed with the way the [DRS 1.3.0 develop branch](https://ga4gh.github.io/data-repository-service-schemas/preview/develop/docs/#tag/Working-With-Compound-Objects/Compound-Objects) frames the need for compound object support:
* Some content (e.g. DICOM images) is best represented as a compound object consisting of a structured collection of atomic DrsObjects.
* Each compound object should have a DRS ID, that clients can use to retrieve the object structure and its constituent atomic objects.
We discussed two possible ways to represent and retrieve compound object contents, but didn’t have time to discuss their tradeoffs:
[...]

PR #406, opened 2024-08-17
Feature/issue 394 drs plus connect docs v1

MichaelLukowski:
See issue [394](https://github.com/ga4gh/data-repository-service-schemas/issues/394)

Here we are proposing Data Connect as a best practice for dealing with linking additional metadata to data objects in DRS


briandoconnor:
In the Cloud WS meeting on [Aug 12th, 2024](https://docs.google.com/document/d/1hayvWLIoymomPI9oXcaTZirn5YxFv1cYAs70zyvlvnA/edit#bookmark=id.jg4lev1aum29) we decided to just add text to the spec that you should have a catalog, such as Data Connect. And have that be sufficient for DRS 1.5.  As a result I'll merge #406.  Note, this merges to master rather than develop which is what should be used.  I'll back port the changes to develop so the release works.

PR #409, opened 2024-11-08
nesting to service-info for stats

MichaelLukowski:
As per @davidlougheed 's comments adding nesting to service-info for DRS related information

### DIFF
--- a/openapi/components/schemas/DrsService.yaml
+++ b/openapi/components/schemas/DrsService.yaml
@@ -6,12 +6,6 @@ properties:
   maxBulkRequestLength:
     type: integer
     description: The max length the bullk request endpoints can handle (>= 1) before generating a 413 error e.g. how long can the arrays bulk_object_ids and bulk_object_access_ids be for this server.  
-  objectCount:
-    type: integer
-    description: The total number of objects in this DRS service.
-  totalObjectSize:
-    type: integer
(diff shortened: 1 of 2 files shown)


davidlougheed:
I'll try to take a look today @MichaelLukowski !

davidlougheed:
first note: I like the change, but this only changes the example, not the OpenAPI schema itself (unless I'm missing something) in https://github.com/ga4gh/data-repository-service-schemas/blob/42a4ba88b83594567139f3c7c66e6f72562607dc/openapi/components/responses/200ServiceInfo.yaml . These need to be harmonized, there is a disconnect between what the OpenAPI schema says and what the example says. this was the case with my original review as well.

MichaelLukowski:
@davidlougheed you're totally right. I just added a commit that addresses your comment

davidlougheed:
looks better @MichaelLukowski ! and it's intentional that `maxBulkRequestLength` is required here? (and a small typo - in "The max length the bullk request endpoints", it should be "bulk" instead)

MichaelLukowski:
hey @davidlougheed, yes the `maxBulkLengh` should be required as that is a part of the Bulk endpoint that is part of DRS 1.4

I'll fix the typo and merge!

davidlougheed:
was curious because in 1.4.0 it didn't show up as required in the browseable API, but I see it is marked as required in `DrsService.yaml` (so I think this is more of a rendering issue).

image

PR #419, opened 2025-12-17
Get by checksum

kellrott:
This adds an additional DRSObject path, to access objects by their checksums.
Being able to access DRSObjects by their checksum is critical for a number of different use cases:

1) A researcher receives a reference from a third party source, and that file has possibly been renamed. They want to see if it has already been catalogued in a DRS server. 
2) Some data management systems (for example [git-lfs](https://git-lfs.com)) use checksums to manage and identify files. Providing a way to identify records by checksum makes it easier to work with these systems (see https://github.com/calypr/git-drs that is currently being developed)

### DIFF
--- a/openapi/components/parameters/Checksum.yaml
+++ b/openapi/components/parameters/Checksum.yaml
@@ -1,0 +1,6 @@
+in: path
+name: checksum
+required: true
+description: A `checksum` value from the `checksums` list of a `DrsObject`
+schema:
+  type: string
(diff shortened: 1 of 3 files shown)


grsr:
Hi @kellrott  - As discussed on slack, I think this PR looks good. The only comments I'd make is whether you think we should add a field to `/service-info` advertising whether the server supports fetching by checksum?

In my upload PR https://github.com/ga4gh/data-repository-service-schemas/pull/418 I have now added support for adding additional checksums to existing objects, with a `PUT` to `/objects/{object_id}/checksums` (and `/objects/checksums` for bulk additions), which I think was something you suggested in a working group call. I think it fits naturally alongside my access method update endpoints, and will complement this PR if they both end up being accepted.

I'll be on the working group call on 11/1/26 if you want to discuss any of this further.

grsr:
Hi @kellrott - As GitHub has no doubt notified you, I have now included the changes in this PR in my #418 PR.

